PR #2, opened 2021-07-28
Transaction are not updated when changing accounts 

shjortConcordium:
## Purpose

Fix the bug that transactions are not updated (fetched from the proxy) when you change accounts.
The bug was that, if we have already finished loading the transactions from the current account,
when we switch accounts, the abortController would be marked as aborted, but would never leave the aborted state.

## Changes

The Abortcontroller now only goes into "aborted" state, if it is already running. (Before it would go into 'aborted' while idle, and would never leave the 'aborted' state)
Removed the onAborted function, finish should always be used when finished, instead of onAborted, when 'aborted' .

## Checklist

      hard-to-understand areas.


orhoj:
Please merge with main so that it runs the build pipeline.

orhoj (approved):

PR #6, opened 2021-07-30
Add ascProvider and logging to notarization

orhoj:
## Purpose
When doing notarization on macOS on a computer with multiple available providers the build fails. This change ensures that it selects the correct input to use.

## Changes
Setting the `ascProvider` and added logging when notarization starts.

## Checklist

      hard-to-understand areas.


Bargsteen (approved):
Looks good :)

PR #14, opened 2021-08-09
Store credential randomness

shjortConcordium:
## Purpose

We have to store the randomness used when creating new credentials.
(Updates concordium-base, and rust now needs to be v1.53)

## Changes

Updated concordium-base dependency and changed rust code to be compatible.
Added randomness to credential table, and save it when creating new credentials.
Updated github actions to use rust v1.53. (instead of current stable).
Changed ChosenAttributesKeys enum to AttributeKey.

## Checklist

      hard-to-understand areas.


soerenbf (approved):
Seems to work as expected 👍

PR #15, opened 2021-08-09
Import resolve naming conflicts

shjortConcordium:
## Purpose

Resolve naming conflicts when importing, as it currently just fails and says the import is incompatible. 

Resolves according to:
Account: use existing, unless name is sub-string of account address. (This is to handle the upcoming account recovery)
Identity: use existing. (When Account Recovery is implemented, this should change name if the identity is a placeholder)
AddressBookEntry name: use existing, unless bound to a local account, which has changed.
AddressBookEntry note: use existing, unless existing is empty.
Credential note: use existing, unless existing is empty.

## Changes

Changes the import functions to handle these conflicts

## Checklist

      hard-to-understand areas.


app/pages/exportImport/PerformImport.tsx
```diff
@@ -38,23 +46,50 @@ import getGenesis from '~/database/GenesisDao';
...
     const [nonDuplicates, duplicates] = partition(entries, (entry) =>
-        hasNoDuplicate(entry, addressBook, addressBookFields, ['note'])
+        hasNoDuplicate(entry, addressBook, addressBookFields)
     );
     if (nonDuplicates.length > 0) {
         await importAddressBookEntry(nonDuplicates);
     }
-    return duplicates;
+    const trueDuplicates = [];
+    if (duplicates.length > 0) {
```
soerenbf, line 64: Isn't this conditional redundant?

app/pages/exportImport/PerformImport.tsx
```diff
@@ -178,16 +267,30 @@ export default function PerformImport({ location }: Props) {
         importedData.accounts
             .filter((account: Account) => account.identityId === identity.id)
             .map((account: Account) => (
-                <p key={account.address}>{account.name}</p>
+                <p key={account.address}>
+                    {account.name}{' '}
+                    <span className="bodyLight">
+                        {messages[account.address] &&
+                            `(${messages[account.address]})`}
+                    </span>
```
soerenbf, line 275: Maybe we should add a different color for the message as well? Also the `span` could be moved to avoid empty elements

```suggestion
                    {messages[account.address] &&
                        <span className="bodyLight textFaded">({messages[account.address]})</span>
                    }
```

app/pages/exportImport/PerformImport.tsx
```diff
@@ -228,16 +331,24 @@ export default function PerformImport({ location }: Props) {
...
                                                     <b>ID:</b> {identity.name}
                                                 </h3>
+                                                <p className="mT0 body3 bodyLight">
+                                                    {messages[identity.id] &&
+                                                        `(${
+                                                            messages[
+                                                                identity.id
+                                                            ]
+                                                        })`}
+                                                </p>
```
soerenbf, line 351: I think this should be on the same line as with account conflict messages for consistency.

app/utils/importHelpers.ts
```diff
@@ -346,24 +452,36 @@ async function importDuplicateWallets(
...
+        if (!identityInDatabase) {
             throw new Error(
                 'Internal error. An existing and matching identity should have been found, but was not.'
             );
         }
 
-        const newIdentityId = newIdentity.id;
+        const newIdentityId = identityInDatabase.id;
+
+        // If there is a naming conflict between the names, ask the user to resolve it.
```
soerenbf, line 473: Does this mean the actual user using the wallet, or isn't this just handled by the resolver function?

soerenbf (commented):
A couple of things I noticed:

- If I import anything that already exists, it reads as if accounts, ID's, etc. (except for address book entries) are imported, even though they are discarded as they already exist? Should we clarify this as well, and not just when a naming conflict appears?
- Took me a while to understand the code. I don't know if it's possible to simplify or split into more pieces? Maybe it isn't and that might also be OK, but there are some really meaty functions in `importHelpers.ts`

app/utils/importHelpers.ts
shjortConcordium: This is an outdated comment, I have changed it to:
> If there is a name conflict, resolve it.

app/pages/exportImport/PerformImport.tsx
shjortConcordium: Yes, I'll remove it.

app/pages/exportImport/PerformImport.tsx
shjortConcordium: I agree, I've added this.

app/pages/exportImport/PerformImport.tsx
shjortConcordium: I've moved it to the same line + added `textFaded` and removed `body3`.

shjortConcordium:
> A couple of things I noticed:
> 
>     * If I import anything that already exists, it reads as if accounts, ID's, etc. (except for address book entries) are imported, even though they are discarded as they already exist? Should we clarify this as well, and not just when a naming conflict appears?
> 
>     * Took me a while to understand the code. I don't know if it's possible to simplify or split into more pieces? Maybe it isn't and that might also be OK, but there are some really meaty functions in `importHelpers.ts`

* I have added messages for duplicates with the same name.
* I have tried to simplify/refactor it.

app/pages/exportImport/PerformImport.tsx
```diff
@@ -228,15 +253,24 @@ export default function PerformImport({ location }: Props) {
...
                                             <div
                                                 key={identity.id}
                                                 className={styles.importSection}
                                             >
-                                                <h3>
-                                                    <b>ID:</b> {identity.name}
+                                                <h3 className="mB0">
+                                                    <b>ID:</b> {identity.name}{' '}
+                                                    {messages[identity.id] && (
+                                                        <span className="bodyLight textFaded body3">
```
soerenbf, line 266: You probably meant to do this but forgot, I assume

```suggestion
                                                        <span className="bodyLight textFaded">
```

app/pages/exportImport/PerformImport.tsx
soerenbf: I think you forgot the part where you wanted to remove `body3`? 😉

soerenbf (commented):
Nice refactor, IMO.

When looking at the imported data now, a couple of things spring to mind:
- I could do with some spacing between the ID line, and the "Accounts" section (added 10px in image below)
- I think it would be nice with a difference in font size and some indication that the messages are an annotation made by the desktop wallet to explain. In the image below I tried adding 2 different ways to do this (I think I prefer the "-" version, but if you prefer another one, I don't mind)

What do you think about these suggestions?

Screenshot 2021-08-13 at 10 53 12

shjortConcordium:
> Nice refactor, IMO.
> 
> When looking at the imported data now, a couple of things spring to mind:
> 
>     * I could do with some spacing between the ID line, and the "Accounts" section (added 10px in image below)
> 
>     * I think it would be nice with a difference in font size and some indication that the messages are an annotation made by the desktop wallet to explain. In the image below I tried adding 2 different ways to do this (I think I prefer the "-" version, but if you prefer another one, I don't mind)
> 
> 
> What do you think about these suggestions?
> 
> Screenshot 2021-08-13 at 10 53 12

I agree with the spacing between identity and account.
I agree we could make it clearer, I prefer the parenthesis though.
I don't think we need to have a different font size, I've tried to add some spacing though:

Screenshot from 2021-08-13 12-54-00

soerenbf:
> I agree with the spacing between identity and account.
> I agree we could make it clearer, I prefer the parenthesis though.
> I don't think we need to have a different font size, I've tried to add some spacing though:

Sounds good to me 👍

soerenbf (approved):

PR #18, opened 2021-08-10
Context isolation

shjortConcordium:
## Purpose

Enable Context Isolation.

## Changes

Enables Context isolation, and moves stuff to preload script.
Communication with database, files and ledger + http and grpc requests are moved to the preload script. 
refactoring of code that use to interface with ipc.

## Checklist

      hard-to-understand areas.


soerenbf (changes requested):
Did regression testing on what i could think of being implicated by the changes, and found that I cannot create an identity (at least using the "concordium test" identity provider) on this branch.

Other essential flows seemed to work fine. Tested the following:
- Simple transfer
- Account creation
- Export/import
- Open application page from menu (terms & conditions)
- Change wallet password

soerenbf (commented):
On startup I get a warning stating that `webview` bypasses the security gains of contextIsolation mode. So maybe we should look into removing webviews as well in favor of `BrowserView`, as you said yourself 😄

shjortConcordium:
> Did regression testing on what i could think of being implicated by the changes, and found that I cannot create an identity (at least using the "concordium test" identity provider) on this branch.
> 
> Other essential flows seemed to work fine. Tested the following:
> 
>     * Simple transfer
> 
>     * Account creation
> 
>     * Export/import
> 
>     * Open application page from menu (terms & conditions)
> 
>     * Change wallet password

I've fixed it. I had to move the http get request back to the main thread.

> On startup I get a warning stating that `webview` bypasses the security gains of contextIsolation mode. So maybe we should look into removing webviews as well in favor of `BrowserView`, as you said yourself smile

I have disabled the webview tag and used Browserview instead.

soerenbf:
> > On startup I get a warning stating that `webview` bypasses the security gains of contextIsolation mode. So maybe we should look into removing webviews as well in favor of `BrowserView`, as you said yourself :smile:
> 
> I have disabled the webview tag and used Browserview instead.

This seems to work now! 👍

soerenbf (approved):
Identity issuance now seems to work as expected as well.

PR #19, opened 2021-08-12
Split keys addressses for ledger comparison

shjortConcordium:
## Purpose

Split display of addresses and keys to match the ledger display, which is made consistent in https://github.com/Concordium/concordium-ledger-app/pull/14.

## Changes

Splits display of keys into chunks of size 16.
Splits display of addresses into chunks of size 10.

## Checklist

      hard-to-understand areas.


app/components/DisplayAddress.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const lineLength = 10;
+
+export default function DisplayAddress({
+    address,
+    lineClassName,
+    outerClassName,
+}: Props) {
+    return (
+        <div className={clsx(outerClassName, 'textCenter')}>
```
orhoj, line 20: I think we should consider making this monospaced to make it easier to validate.
```
4H4QJaLRQv
p8mzcDCxVu
11scJCcrvt
Xnd1fhwjSw
HTDTHcXuD4
```

app/components/ledger/CredentialInfoLedgerDetails.tsx
```diff
@@ -11,13 +13,17 @@ const CredentialInfoLedgerDetails = (props: Props) => (
...
+                publicKey={props.credentialPublicKeys.keys[0].verifyKey}
+            />
         </p>
         <p>
             <b>Signature threshold:</b> {props.credentialPublicKeys.threshold}
         </p>
         <p>
-            <b>Registration ID credential (RegIdCred):</b> {props.credId}
+            <b>Registration ID credential (RegIdCred):</b>
+            <PublicKeyDetails publicKey={props.credId} />
```
orhoj, line 26: It is not a public key you are supplying here. Perhaps it would be better to make `PublicKeyDetails` into a more generic component for displaying hex strings, and then make `PublicKeyDetails` use that.

orhoj:
Should the updated formatting be used in the address book? In the account address view for an account? I am not sure if we should _always_ display it on multiple lines, or if we should only do it when submitting to the Ledger.

shjortConcordium:
> Should the updated formatting be used in the address book? In the account address view for an account? I am not sure if we should _always_ display it on multiple lines, or if we should only do it when submitting to the Ledger.

I think we will only use this 5 line splitting when submitting to the ledger.

orhoj (approved):

PR #20, opened 2021-08-13
Notifications

soerenbf:
## Purpose

To add a less intrusive way (than modals) to show notifications to the user. They can be automatically removed after a given timeout, or closed manually. Will also be used for displaying an "update" notification.

## Changes

- Added a redux slice for notifications to live in
- Added components `Notification` and `Notifications`
- Renders notifications in a fixed position by calling the `pushNotification` function from `NotificationSlice`

## Checklist

      hard-to-understand areas.

## Testing

To test, call the `pushNotification` function with different notification types. I used `setInterval` myself to have it display over an over.


app/features/NotificationSlice.ts
```diff
@@ -0,0 +1,85 @@
...
+export const { triggerUpdateNotification, removeNotification } = actions;
+
+/**
+ * Display notifications of different types.
+ *
+ * @example
+ * const dispatch = useDispatch();
+ *
+ * pushNotification(dispatch, { NotificationLevel.Info, message: 'Informative message' }, 5);
+ * pushNotification(dispatch, { NotificationLevel.Update });
```
shjortConcordium, line 64: You can't add an update notification with pushNotification, because of the types.
Either remove this from the example or update the types to allow it.

shjortConcordium (commented):
Screenshot from 2021-08-13 13-20-52

I think it could be nice with a little bit of vertical space between the notifications.

app/features/NotificationSlice.ts
soerenbf: True, I removed it as the update notification is meant to be triggered through the exported action instead.

soerenbf:
> I think it could be nice with a little bit of vertical space between the notifications.

Added spacing between.

shjortConcordium (approved):

shjortConcordium (approved):

PR #22, opened 2021-08-13
Cleanup failed/rejected transation changes in account report

shjortConcordium:
## Purpose

Fix that, in the account report, failed/rejected transactions have incorrect values in the "public/shielded balance change" columns, currently.

## Changes

Changes the values in "public/shielded balance change" columns for failed and rejected transactions in the account report.

## Checklist

      hard-to-understand areas.


app/pages/Accounts/AccountReport/util.ts
```diff
@@ -41,6 +51,13 @@ function calculateShieldedBalanceChange(
     transaction: TransferTransaction,
     isOutgoing: boolean
 ): string {
+    if (
+        [TransactionStatus.Failed, TransactionStatus.Rejected].includes(
+            transaction.status
+        )
+    ) {
+        return '0';
+    }
```
soerenbf, line 60: I ændringen før denne, er der forskel på "rejected" og "failed" ift. om transaction cost bliver inkluderet. Hvordan kan det være, at det ikke gør sig gældende her?

shjortConcordium: Failed transactions are those, which are added to the chain, but "failed", so the user still pay the cost, but the transfer doesn't happen. Rejected transactions do not make it on chain, so they have no effect.

That is why they are handled different on the public balance, but the cost does not affect the shielded balance.

soerenbf (approved):

PR #23, opened 2021-08-13
Remove failed identity

shjortConcordium:
## Purpose

Currently a failed identity cannot be removed from the wallet. This adds the ability to do so.

## Changes

When identity fails, set initial account status to failed instead of deleting it.
Identity failed modal refactored out of IdentityView, and now appears, when clicking on failed initial accounts.
Identity failed modal now removes the identity and its initial account.
Pressing "try again", on failed modal will send you to the identity issuance with the names initialized. 

## Checklist

      hard-to-understand areas.


app/components/FailedIdentityModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }, [isRejected, setModalOpen]);
+
+    async function remove() {
+        await removeIdentityAndInitialAccount(dispatch, identityId);
+        setModalOpen(false);
+    }
+
+    return (
+        <ChoiceModal
```
soerenbf, line 42: On the `Modal` component, it's possible to set a `disableClose` prop, which I think would fit this use case. When the close button is there, it's possible for users to opt out of removing the identity/initial account.

app/pages/IdentityIssuance/IdentityIssuancePage.tsx
```diff
@@ -33,18 +33,27 @@ function getSubtitle(location: string) {
...
     const [provider, setProvider] = useState<IdentityProvider | undefined>();
-    const [initialAccountName, setInitialAccountName] = useState<string>('');
-    const [identityName, setIdentityName] = useState<string>('');
-    const { pathname } = useLocation();
+    const [initialAccountName, setInitialAccountName] = useState<string>(
+        state?.identityName || ''
+    );
+    const [identityName, setIdentityName] = useState<string>(
+        state?.initialAccountName || ''
+    );
```
soerenbf, line 56: You're using `identityName` for `initialAccountName` and vice versa 😆

app/features/IdentitySlice.ts
```diff
@@ -58,6 +71,15 @@ export async function loadIdentities(dispatch: Dispatch) {
     dispatch(updateIdentities(identities));
 }
 
+export async function removeIdentityAndInitialAccount(
+    dispatch: Dispatch,
+    identityId: number
+) {
+    await removeIdentityAndInitialAccountInDatabase(identityId);
+    await loadAccounts(dispatch);
+    return dispatch(removeIdentityInRedux(identityId));
```
soerenbf, line 80: I know we do this other places than this, and that removing an item from a list is fairly simple, but I'm not a fan of the duplication happening when we first change something in the DB, and afterwards replicate the change in redux as well.

Performance-wise, I know this is probably the best way, but it would be way simpler if we could have a general model for how updates are done to the DB and redux that relies on a single source for data. When we work with a local DB anyway, I don't think the performance gain is worth the added complexity (again, this action isn't complex).

What are your thoughts on this?

soerenbf (commented):
A couple of things and thoughts in the comments below

app/features/IdentitySlice.ts
shjortConcordium: I think you are correct, the performance gain does not justify adding the potential of de-synchronization between the db and redux.

app/components/FailedIdentityModal.tsx
shjortConcordium: I've used this now, as pressing 'later' does the same.

shjortConcordium:
I've tried to make adjustments, like we discussed.
I don't know if I like adding a status type to keep track of which identities have been warned about. (but I can't think of a solution I prefer)

app/features/IdentitySlice.ts
```diff
@@ -58,6 +61,15 @@ export async function loadIdentities(dispatch: Dispatch) {
     dispatch(updateIdentities(identities));
 }
 
+export async function removeIdentityAndInitialAccount(
+    dispatch: Dispatch,
+    identityId: number
+) {
+    await removeIdentityAndInitialAccountInDatabase(identityId);
+    await loadAccounts(dispatch);
+    return loadIdentities(dispatch);
```
soerenbf, line 70: This in the nitpicky area, but as the 2 are decoupled, you could do:

```suggestion
    return Promise.all([loadAccounts(dispatch), loadIdentities(dispatch)]);
```

soerenbf (approved):
Looks good! Only 1 small comment. If you think it's irrelevant, you can just resolve it.

PR #26, opened 2021-08-18
Show node connection

shjortConcordium:
## Purpose

Add a node status in the sidebar to see its status. (And whether it is catching up)
Current states:
pinging, ready, catching up, unavailable.

## Changes

Adds a node status component to the side bar.
Adds getPeerList grpc call.
Updates pending icon.

## Checklist

      hard-to-understand areas.


app/shell/Status/Status.module.scss
```diff
@@ -1,21 +1,33 @@
 @import 'config/bundle';
 
+.nodeStatus,
 .body {
     composes: body4 from global;
     display: flex;
     align-items: center;
     justify-content: center;
     margin: 20px;
```
soerenbf, line 9: For me, the spacing between statuses looks a little off. Following suggestion looked better to me.

```suggestion
    margin: 10px 20px;
```

What do you think?

app/shell/Status/NodeStatus.tsx
```diff
@@ -0,0 +1,81 @@
...
+            const upToDate = await isNodeUpToDate();
+            status = upToDate ? Status.Ready : Status.CatchingUp;
+        } catch {
+            // do nothing, status defaults to unavailable.
+        } finally {
+            if (!controller.isAborted) {
+                setStatusText(status);
+                setTimeout(setStatus, checkInterval, controller);
+            }
+        }
```
soerenbf, line 59: Setting the status text to `Status.Pinging` with the recurring timeout causes the UI to flicker for a moment while waiting for the grpc response. I think it's better to assume the status doesn't change and only change the status when it does.

app/shell/Status/NodeStatus.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }, []);
+
+    useEffect(() => {
+        if (setStatus) {
+            const controller = new AbortController();
+            setStatus(controller);
+            return () => controller.abort();
+        }
+        return noOp;
+    }, [connectionSettings?.value, setStatus]);
```
soerenbf, line 69: Instead of using abortcontroller for this, wouldn't it make more sense (semantically, but also in terms of keeping it simple) to do:

```suggestion
    useEffect(() => {
        const interval = setInterval(setStatus, checkInterval);
        return () => clearInterval(interval);
    }, [connectionSettings?.value, setStatus]);
```

And then of course adapt `setStatus` accordingly?

soerenbf (commented):
I think we should align how the statuses look, including:

- Text alignment (wallet is center, node is left)
- Icons (none on wallet, node has icons)
- Text colors (wallet uses text colors to display status instead of icons, node seems to not use text colors)

I think we should run this by Jens.

app/shell/Status/NodeStatus.tsx
```diff
@@ -0,0 +1,81 @@
...
+        } finally {
+            if (!controller.isAborted) {
+                setStatusText(status);
+                setTimeout(setStatus, checkInterval, controller);
+            }
+        }
+    }, []);
+
+    useEffect(() => {
+        if (setStatus) {
```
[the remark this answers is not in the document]
shjortConcordium: Never, it seems.

app/shell/Status/Status.module.scss
shjortConcordium: I agree.

app/shell/Status/NodeStatus.tsx
shjortConcordium: I've delayed showing pinging to a second after waiting.

shjortConcordium:
> I think we should align how the statuses look, including:
> 
>     * Text alignment (wallet is center, node is left)
> 
>     * Icons (none on wallet, node has icons)
> 
>     * Text colors (wallet uses text colors to display status instead of icons, node seems to not use text colors)
> 
> 
> Screenshot 2021-08-23 at 09 13 23
> 
> I think we should run this by Jens.

I have updated it, following Jens' sketches.

PR #29, opened 2021-08-24
Add logging

shjortConcordium:
## Purpose

Add logging

## Changes

Added logging capabilities.
Log errors throughout the app.
Export the logs from the application menu.

## Checklist

      hard-to-understand areas.


soerenbf:
Logging for auto updates needs to be changed to use new logging implementation as well.

soerenbf (approved):
Looks good!

orhoj:
It seems like that the logging does not provide the filename (and perhaps also the line number) of where it was logged from. This is required to make it useful to debug from.

orhoj:
I am a bit uncertain about what to do when logging errors caught in a try-catch, as the object thrown can have any type, so I am not sure if what we have here works in general, or will break down in certain cases?

shjortConcordium:
> I am a bit uncertain about what to do when logging errors caught in a try-catch, as the object thrown can have any type, so I am not sure if what we have here works in general, or will break down in certain cases?

Well we just include the error in the log, so as long as the object is serializable, then it should work perfectly, and otherwise those "non-serializable" details will just not be in the log.

PR #40, opened 2021-09-01
Automatic updates for mainnet

soerenbf:
## Purpose

To bring automatic updates to the application. This is done by publishing new releases to GitHub releases alongside normal distribution channels.

This is only enabled for, and only works for mainnet releases.

## Changes

- Automatic update notification will trigger when an update is available for the application
- Update is verified by matching checksum and signature of downloaded binary with published checksum and signature
  - Assumes public key for verification will be accessible at: "https://distribution.mainnet.concordium.com/tools/concordium-desktop-wallet-pubkey.pem"
- Script for generating verification assets (sha256 checksum + signature) of binaries
- Updated readme with description of how to use the aforementioned script

## Checklist

      hard-to-understand areas.


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 -   Now shows connected node status in side bar.
 -   It is now possible to remove a failed identity.
 -   Added reference (sessionId) on failed identities, for support.
+-   Application will now update automatically (Windows, MacOS, and AppImage for Linux)
```
orhoj, line 15: Application will automatically search for updates and prompt the user to download and install them (Windows, macOS and AppImage for Linux).

app/components/AppUpdateNotification.tsx
```diff
@@ -0,0 +1,154 @@
...
+    const update = useCallback(() => {
+        onUpdate();
+        setState({ status: UpdateStatus.Downloading });
+    }, [onUpdate, setState]);
+
+    return (
+        <>
+            <div className="flex alignCenter">
+                <CogIcon width="21" height="22" />
+                <span className="mL5">New updates are available</span>
```
orhoj, line 53: Perhaps it could state the version that is available like:
```
Version 1.2.0 is available
```

app/main/autoUpdate.ts
```diff
@@ -0,0 +1,143 @@
...
+    return new Promise((resolve, reject) => {
+        stream.on('end', () => resolve(verifier));
+        stream.on('error', reject);
+    });
+}
+
+const releasesFeed = `https://github.com/${build.publish.owner}/${build.publish.repo}/releases/download`;
+
+function getVerificationFunctions(
+    { downloadedFile: filePath, releaseName, path: fileName }: RealUpdateInfo // UpdateInfo interface doesn't seem to be aligned with actual content.
```
orhoj, line 79: The comment should be moved to the interface definition instead.

app/main/autoUpdate.ts
```diff
@@ -0,0 +1,143 @@
...
+) {
+    const releaseFileUrl = `${releasesFeed}/${releaseName}/${fileName}`;
+
+    return {
+        async verifyChecksum() {
+            const remoteHash = (
+                await getRemoteContent(`${releaseFileUrl}.hash`)
```
orhoj, line 86: I think we should use a file format extension like `.sha256` or `.sha256sum` to indicate what type of hash is contained in the file.

app/main/autoUpdate.ts
```diff
@@ -0,0 +1,143 @@
...
+    stream.on('data', (data) => sum.update(data));
+
+    return new Promise((resolve, reject) => {
+        stream.on('end', () => resolve(sum.digest('hex')));
+        stream.on('error', reject);
+    });
+}
+
+function getVerifier(path: string): Promise<Verify> {
+    const verifier = createVerify('RSA-SHA256');
```
orhoj, line 65: We should use `ed25519` instead of RSA for making the signature. So you need to change the verification to verify here. Elsewhere we verify signatures using `noble-ed25519` in the desktop wallet, but perhaps crypto can also do it.

app/main/autoUpdate.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    const { verifyChecksum, verifySignature } = getVerificationFunctions(
+        info as RealUpdateInfo
+    );
+
+    try {
+        await Promise.all([verifyChecksum(), verifySignature()]);
+        mainWindow.webContents.send(updateVerified);
+    } catch (e) {
+        log.error('Could not update application due to:', e);
```
orhoj, line 128: Does this way of logging align with the logging branch that is open elsewhere? The logs from the auto updater should also end up in a log file to be useful to us.

app/main/autoUpdate.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    return new Promise((resolve, reject) => {
+        stream.on('end', () => resolve(verifier));
+        stream.on('error', reject);
+    });
+}
+
+const releasesFeed = `https://github.com/${build.publish.owner}/${build.publish.repo}/releases/download`;
+
+function getVerificationFunctions(
```
orhoj, line 78: Caro suggests that ensure that this is covered by automatic tests. This should be possible by mocking `axios` to return the required sha256 checksum and signature for each case to be tested.

app/components/AppUpdateNotification.tsx
```diff
@@ -0,0 +1,154 @@
...
+            {status === UpdateStatus.Downloading && 'Downloading update'}
+            {status === UpdateStatus.Verifying && 'Verifying update'}
+        </div>
+    </>
+);
+
+const SuccessChild = ({ onPostpone }: ChildProps) => {
+    function handleEnd() {
+        window.autoUpdate.quitAndInstall();
+        onPostpone();
```
orhoj, line 82: I do not understand why `onPostpone()` is called when we succeeded with the update.

scripts/fileDigest.js
```diff
@@ -0,0 +1,265 @@
...
+    linux,
+    dir,
+    appVersion,
+} = argv;
+
+const hashAlgorithm = 'sha256';
+
+// Produce file checksum
+async function writeChecksum(file) {
+    const { stdout } = await exec(`openssl dgst -${hashAlgorithm} ${file}`);
```
orhoj, line 88: As we are already requiring `node` we should just use that for hashing and verification of signatures instead of adding the requirement of having `openssl` available as well.

scripts/fileDigest.js
```diff
@@ -0,0 +1,265 @@
...
+ */
+async function verifySignature(pubKeyPath, file, sigFile) {
+    const { stdout } = await promisify(exec)(
+        `openssl dgst -${hashAlgorithm} -verify ${pubKeyPath} -signature ${sigFile} ${file}`
+    );
+
+    console.log(stdout);
+}
+
+function getPublicKey() {
```
orhoj, line 111: It would make sense to re-use the code you already have for getting the public key elsewhere, instead of having two implementations for it.

scripts/fileDigest.js
```diff
@@ -0,0 +1,265 @@
...
+    if (verifyKeyPath) {
+        verifySignature(verifyKeyPath, file, sigOutFile);
+    } else if (!skiprv) {
+        verifyRemote(file, sigOutFile);
+    }
+}
+
+const linuxTargets = build.linux.target;
+const winTargets = ['exe'];
+const macTargets = ['dmg', 'zip'];
```
orhoj, line 218: I do not believe `zip` should be listed here.

scripts/fileDigest.js
```diff
@@ -0,0 +1,265 @@
...
+(async () => {
+    if (!filesToDigest.length) {
+        console.error(
+            'Found no files for specified runtime arguments in directory:',
+            dir
+        );
+    }
+    for (let i = 0; i < filesToDigest.length; i += 1) {
+        const file = filesToDigest[i];
+        const shouldVerify = i === filesToDigest.length - 1;
```
orhoj, line 252: This means that we only verify the final file?

app/main.ts
```diff
@@ -156,9 +145,9 @@ const createWindow = async () => {
 
     initializeIpcHandlers(mainWindow, printWindow, browserView);
 
-    // Remove this if your app does not use auto updates
-    // eslint-disable-next-line
-    new AppUpdater();
+    if (process.env.NODE_ENV === 'production' || process.env.TARGET_NET) {
```
orhoj, line 148: If in development mode and a `TARGET_NET` has been specified, why should it then run the auto updater?

README.md
```diff
@@ -128,3 +136,19 @@ The notarizing process can take a while, and for testing purposes it can be skip
...
+
+To generate the assets, run:
+
+```bash
+yarn generate-verification-assets <path-to-private-key>
+```
+
+-   The private key used for this needs to match the public key published for verification, otherwise verification will not succeed.
+
+By default, the script tries to generate a hash and sig file for all release binaries in `release/`. It also tries to verify the signature created with a public key, which is published separately.
```
orhoj, line 152: Change to:
```
It also tries to verify the created signatures with a public key, which is published separately.
```

app/main/autoUpdate.ts
```diff
@@ -0,0 +1,143 @@
...
+    const { verifyChecksum, verifySignature } = getVerificationFunctions(
+        info as RealUpdateInfo
+    );
+
+    try {
+        await Promise.all([verifyChecksum(), verifySignature()]);
+        mainWindow.webContents.send(updateVerified);
+    } catch (e) {
+        log.error('Could not update application due to:', e);
+        mainWindow.webContents.send(updateError, 'Verification failed');
```
orhoj, line 129: After talking to Caro we came to the conclusion that this message should probably be "eased" a bit, so that the user does not panic about something having failed, as it will quite likely be because we uploaded a release incorrectly. Perhaps we can come up with something together with Jens so that the user just carries on with their work, and then they will just get the update sometime in the future.

orhoj:
Remember to create a documentation issue to make sure that our user documentation is updated accordingly.

CHANGELOG.md
soerenbf: I agree

app/components/AppUpdateNotification.tsx
soerenbf: It's bad naming. The event is only used to close the notification. I'll rename

scripts/fileDigest.js
soerenbf: Yes, I agree.

app/main.ts
soerenbf: Doesn't make any sense.. Changed to `&&` instead.

scripts/fileDigest.js
soerenbf: Yes, the assumption is, that because everything is signed in one batch with the same key, it's redundant to verify every file as the verification is only used to prevent errors with uploading incorrectly signed binaries.

app/main/autoUpdate.ts
soerenbf: On this branch `electron-log` is used, but I agree that we should align. I talked this through with @shjortConcordium, and he suggested that he'd adapt in his branch if it's merged later. Otherwise if we get logging merged before, I'll include it in this PR.

scripts/fileDigest.js
soerenbf: It would, but this script is not compiled and can't use esmodules, so it gets hard to reuse code used in the app. The public key location is reused.

app/main/autoUpdate.ts
soerenbf: I adapted the signing and verification functionality to work with keys generated with `ed25519`.

app/components/AppUpdateNotification.tsx
soerenbf: Talked it over with @jens-concordium as well, and we agreed that version number could be useful to know in different scenarios. It is now included.

app/main/autoUpdate.ts
soerenbf: I agree, we've decided on "Could now apply update."

app/main/autoUpdate.ts
soerenbf: I've added tests for this.

app/components/AppUpdateNotification.tsx
orhoj: This is still called `onPostpone()` here.

scripts/fileDigest.js
orhoj: I don't see any reason for this optimization, so I believe the check should be made for every file.

app/components/AppUpdateNotification.tsx
soerenbf: Fixed

scripts/fileDigest.js
soerenbf: Alright, I removed it.

app/components/AppUpdateNotification.tsx
```diff
@@ -0,0 +1,160 @@
...
+    const update = useCallback(() => {
+        onUpdate();
+        setState({ status: UpdateStatus.Downloading });
+    }, [onUpdate, setState]);
+
+    return (
+        <>
+            <div className="flex alignCenter">
+                <CogIcon width="21" height="22" />
+                <span className="mL5">Version {version} is available.</span>
```
orhoj, line 59: In my testing this just shows as:
```
Version is available.
```
i.e. `{version}` seems to be undefined.

app/components/AppUpdateNotification.tsx
```diff
@@ -0,0 +1,160 @@
...
+    );
+};
+
+const iconWidth = 60;
+
+const ProcessChild = ({ status }: ChildProps) => (
+    <>
+        <Loading inline iconWidth={iconWidth} />
+        <div className="mT10 textCenter">
+            {status === UpdateStatus.Downloading && 'Downloading update'}
```
orhoj, line 79: While testing I tried to mess up some of the naming/versioning so they mismatched. In that case it cannot find the update:
```
Error: Error: Cannot download "https://github.com/orhoj/concordium-desktop-wallet/releases/download/v1.2.0-alpha2/concordium-desktop-wallet-1.2.0-alpha2.AppImage", status 404:
```
and the UI keeps saying `Downloading update`. It should update the status in this case instead.

app/components/AppUpdateNotification.tsx
soerenbf: should be fixed now.. thought I had checked this - sorry! 😅

app/components/AppUpdateNotification.tsx
soerenbf: good catch, added error handling which should fix this.

orhoj (approved):

PR #16, opened 2021-08-09
Enforce minimum ledger version

shjortConcordium:
## Purpose

Add the functionality to enforce a minimum version for the app on the ledger device.

## Changes

Adds an "outdated" state for the ledger, which asks the user to update the application, and blocks the usage of it.
\+ some minor refactoring in `ledgerObserverImpl.ts` .

## Checklist

- [x] My code follows the style of this project.
- [x] The code compiles without warnings.
- [x] I have performed a self-review of the changes.
- [x] I have documented my code, in particular the intent of the
      hard-to-understand areas.
- [] (If necessary) I have updated the CHANGELOG.

PR #44, opened 2021-09-06
Memo transactions

shjortConcordium:
## Purpose

Add support for memo transaction.
Pairs with changes in https://github.com/Concordium/concordium-ledger-app/pull/16.

## Changes

Allow sending transactions with memos. (simple, encrypted and scheduled)
Added memos to the transaction type and in the database.
Load transactions from the wallet-proxy's v1 path. (instead of v0)
Display memos in transaction view.

## Checklist

      hard-to-understand areas.


app/utils/types.ts
```diff
@@ -187,10 +189,15 @@ export enum TransactionKindId {
...
     Register_data = 21,
+    Simple_transfer_with_memo = 22,
+    Encrypted_transfer_with_memo = 23,
+    Transfer_with_schedule_and_memo = 24,
 }
+
 export interface SimpleTransferPayload {
     amount: string;
     toAddress: string;
+    memo?: string;
```
orhoj, line 200: The typing is better if you make a new interface:
```
export interface SimpleTransferWithMemoPayload extends SimpleTransferPayload {
    memo: string;
}
```
Leaving it as optional on the regular simple transfer is not correct as it cannot have a memo.

orhoj: And after writing this I saw `SimpleTransferWithMemo`, but then I don't understand why you need an optional `memo` here.

app/utils/types.ts
```diff
@@ -786,7 +803,17 @@ export function instanceOfAccountTransactionWithSignature(
 export function instanceOfSimpleTransfer(
```
orhoj, line 803: I think the design is better if we keep the two transactions separate. So I would not expect `instanceOfSimpleTransfer` to match for one with a memo.

scripts/cbor.ts
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable no-console */
+import cbor from 'cbor';
```
orhoj, line 2: This file should be deleted.

package.json
```diff
@@ -40,6 +40,7 @@
     "test-e2e": "node ./internals/scripts/CheckBuildsExist.js && cross-env NODE_ENV=test testcafe electron:./app ./test/e2e/HomePage.e2e.ts",
     "test-e2e-live": "node ./internals/scripts/CheckBuildsExist.js && cross-env NODE_ENV=test testcafe --live electron:./app ./test/e2e/HomePage.e2e.ts",
     "test-watch": "yarn test --watch",
+    "cbor": "ts-node --project scripts/migrations/tsconfig.json -r tsconfig-paths/register ./scripts/cbor.ts",
```
orhoj, line 43: Delete?

app/utils/transactionSerialization.ts
```diff
@@ -34,28 +34,42 @@ import {
...
+    const amount = encodeWord64(BigInt(payload.amount));
+    return Buffer.concat([kind, address, amount]);
+}
 
-    serialized[0] = TransactionKind.Simple_transfer;
-    putBase58Check(serialized, 1, payload.toAddress);
-    put(serialized, 32 + 1, encodeWord64(BigInt(payload.amount)));
-    return Buffer.from(serialized);
+function serializeSimpleTransferWithMemo(payload: SimpleTransferPayload) {
+    if (!payload.memo) {
```
orhoj, line 55: It should not be necessary to have this check here if we type the input parameter better.

app/utils/transactionHelpers.ts
```diff
@@ -592,13 +613,36 @@ export function validateBakerStake(
...
 }
 
+export function validateMemo(memo: string): string | undefined {
+    const asNumber = Number(memo);
+    if (
+        Number.isInteger(asNumber) &&
+        (asNumber > Number.MAX_SAFE_INTEGER ||
+            asNumber < Number.MIN_SAFE_INTEGER)
+    ) {
+        return `Only numbers up to ${Number.MAX_SAFE_INTEGER} is supported.`;
```
orhoj, line 623: I suggest replacing the above with:
```
Numbers greater than 9007199254740991 or smaller than -9007199254740991 are not supported.
```
If there is a space concern and you want to keep the current, then you need to change _is_ to _are_.

app/utils/transactionHelpers.ts
```diff
@@ -592,13 +613,36 @@ export function validateBakerStake(
...
+    }
+    if (getEncodedSize(memo) > maxMemoSize) {
+        return `Memo is too large, encoded size must be at most ${maxMemoSize} bytes`;
+    }
+    // Check that the memo only contains ascii characters
+    // eslint-disable-next-line no-control-regex
+    if (/[^\u0000-\u007f]/.test(memo)) {
```
orhoj, line 630: This deserves its own utility function as it is a candidate for reuse. So refactor it into an `isASCII` method.

app/utils/transactionHandlers/SimpleTransferHandler.tsx
```diff
@@ -55,6 +55,7 @@ export default class SimpleTransferHandler
         amount,
         recipient,
         signatureAmount,
+        memo,
```
orhoj, line 58: The memo transactions should, in my opinion, have their own separate transaction handlers.

app/utils/transactionCosts.ts
```diff
@@ -75,7 +76,10 @@ export function calculateCost(
     );
 }
 
-function getPayloadSizeEstimate(transactionKind: TransactionKindId) {
+/**
+   Note that this function does not support scheduled transfers nor transfers with memos.
```
orhoj, line 80: It would be good to document what the function does now that you have written what it _does  not_.

app/utils/transactionCosts.ts
```diff
@@ -125,24 +131,38 @@ function getEnergyCostOfType(transactionKind: TransactionKindId) {
     }
 }
 
-export function getScheduledTransferPayloadSize(scheduleLength: number) {
+export function getScheduledTransferPayloadSize(
+    scheduleLength: number,
+    memoLength?: number
```
orhoj, line 136: This is never called without `memoLength` being set.

app/utils/transactionCosts.ts
```diff
@@ -125,24 +131,38 @@ function getEnergyCostOfType(transactionKind: TransactionKindId) {
...
+        scheduleLength * 16 +
+        // Memo
+        (memoLength ? 2 + memoLength : 0)
+    );
 }
 
 /**
  *  Given the signatureAmount and schedule length,
  * returns the energy cost of a scheduled transfer.
+ * Param memoLength
```
orhoj, line 155: Use @param to document input parameters.

app/utils/cborHelper.ts
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable import/prefer-default-export */
```
orhoj, line 1: I don't think you need this eslint-disable anymore.

app/utils/cborHelper.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Given a string or number, return a buffer containing the value under cbor encoding.
+ * N.B. given a string, this will attempt to convert it to a number, to decrease the encoded size.
+ */
+export function encodeAsCBOR(value: string | number): Buffer {
+    // Prefer saving as numbers:
+    const asNumber = Number(value);
+    if (Number.isInteger(asNumber)) {
+        if (
+            asNumber >= Number.MAX_SAFE_INTEGER ||
```
orhoj, line 14: `Number.MAX_SAFE_INTEGER` is also safe, so the equality should be `>` and `<` respectively.

app/utils/cborHelper.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+    return Buffer.from(encode(value));
+}
+
+/**
+ * Decode cbot encoding.
+ * @param value is assumed to be a Hex string, containing cbor encoded bytes.
+ */
+export function decodeCBOR(value: string) {
+    return decode(Buffer.from(value, 'hex'));
```
orhoj, line 29: This seems to return an `any` type, and elsewhere we `toString()` it. What happens if the `memo` we try to decode is not `CBOR`, or if it is not something that `toString()` works nicely on?

orhoj:
Very large integer input for a memo crashes the application.

app/pages/multisig/AccountTransactions/CreateTransferProposal/CreateTransferProposal.tsx
```diff
@@ -79,14 +82,20 @@ function CreateTransferProposal({
 }: Props): JSX.Element {
     const dispatch = useDispatch();
 
+    const consensusStatus = useConsensusStatus();
+    const allowMemo = consensusStatus && consensusStatus.protocolVersion > 1;
```
orhoj, line 86: Refactor into a shared method (as this is used in multiple places) calling something like `nodeSupportsMemo()`.

app/pages/Accounts/TransactionListElement.tsx
```diff
@@ -231,6 +275,7 @@ interface Props {
 function TransactionListElement({
     transaction,
     onClick,
+    showFullMessage = false,
```
orhoj, line 278: Should this be `showFullMemo`?

app/pages/Accounts/TransactionListElement.tsx
```diff
@@ -214,6 +221,42 @@ function statusSymbol(status: TransactionStatus) {
     }
 }
 
+const memoTypes = [
+    TransactionKindString.Transfer,
```
orhoj, line 225: It is confusing that `Transfer`, `EncryptedAmountTransfer` and `TransferWithSchedule` are part of `memoTypes` as they are not memo types.

app/pages/Accounts/TransactionListElement.tsx
```diff
@@ -214,6 +221,42 @@ function statusSymbol(status: TransactionStatus) {
...
+
+function showMemo(
+    memo: string | undefined,
+    showFullMessage: boolean,
+    transactionKind: TransactionKindString
+) {
+    if (!memo && showFullMessage && memoTypes.includes(transactionKind)) {
+        return (
+            <i className="body4 m0 mT5 textFaded">
+                The transaction contains no message
```
orhoj, line 241: `message` -> `memo`.

app/features/ledger/Transfer.ts
```diff
@@ -57,6 +60,38 @@ const INS_REMOVE_BAKER = 0x14;
...
+type Memo = string;
+
+/**
+ * Send memo to ledger for signing.
+ * N.B. This does not send the length of the memo, and the CBOR encoded string.
+ */
+async function sendMemo(
+    transport: Transport,
+    ins: number,
+    memo: Memo | undefined,
```
orhoj, line 72: If we type thing correctly, then this should not allow for an undefined value as far as I see it.

app/features/ledger/Transfer.ts
```diff
@@ -57,6 +60,38 @@ const INS_REMOVE_BAKER = 0x14;
...
+    if (!memo) {
+        throw new Error('Unexpected Missing memo');
+    }
+
+    const encodedMemo = encodeAsCBOR(memo);
+
+    let response;
+    const chunks = chunkBuffer(encodedMemo, 255);
+    for (const chunk of chunks) {
+        // eslint-disable-next-line  no-await-in-loop
```
orhoj, line 85: I thought we had disabled the check for `no-await-in-loop` globally?

app/features/ledger/Transfer.ts
```diff
@@ -57,6 +60,38 @@ const INS_REMOVE_BAKER = 0x14;
...
+        response = await transport.send(0xe0, ins, p1, p2, Buffer.from(chunk));
+    }
+    if (!response) {
+        throw new Error('Unexpected missing response from ledger;');
+    }
+    const signature = response.slice(0, 64);
```
orhoj, line 91: Else you refactored this from
```
    const signature = response.slice(0, 64);
    return signature;
```
to
```
    return response.slice(0, 64);
```
Pick one and stay consistent.

jest.config.js
```diff
@@ -10,7 +10,10 @@ module.exports = {
     },
     moduleFileExtensions: ['js', 'jsx', 'ts', 'tsx', 'json'],
     moduleDirectories: ['node_modules', 'app/node_modules'],
-    setupFiles: ['./internals/scripts/CheckBuildsExist.js'],
+    setupFiles: [
+        './internals/scripts/CheckBuildsExist.js',
+        './test/loadModules.ts',
```
orhoj, line 15: Why is this needed?

app/utils/transactionCosts.ts
shjortConcordium: Some tests did, but I changed that.

app/utils/cborHelper.ts
shjortConcordium: It seems that it crashes.

shjortConcordium:
> Very large integer input for a memo crashes the application.

 Yes, it is because the encodedSize is calculated (to calculate cost) while inputting, and I made the encode function crash if given a very large integer.
 I have changed it getEncodedSize catches the error and returns 0 instead.

app/pages/Accounts/TransactionListElement.tsx
shjortConcordium: I have refactored the variable away.

app/features/ledger/Transfer.ts
shjortConcordium: I have refactored it for all Transfers.

jest.config.js
shjortConcordium: Because otherwise we get a ReferenceError on TextDecoder on test files that import the cbor dependency.

app/utils/types.ts
shjortConcordium: I have refactored it to make the transaction more separate.

app/utils/cborHelper.ts
shjortConcordium: I have handled this in TransactionConverters (the only place this is used). It tries to decode, and otherwise saves the Hex string directly.

app/utils/transactionHelpers.ts
```diff
@@ -592,13 +665,35 @@ export function validateBakerStake(
...
 }
 
+export function validateMemo(memo: string): string | undefined {
+    const asNumber = Number(memo);
+    if (
+        Number.isInteger(asNumber) &&
+        (asNumber > Number.MAX_SAFE_INTEGER ||
+            asNumber < Number.MIN_SAFE_INTEGER)
+    ) {
+        return `Numbers greater than ${Number.MAX_SAFE_INTEGER} or smaller than ${Number.MIN_SAFE_INTEGER} are not supported..`;
```
orhoj, line 675: Too many `.`.

orhoj:
When receiving a transaction with a memo the memo does not show.

orhoj:
Receiving an encrypted transfer with a memo does not work correctly.

orhoj:
After merge with the account recovery branch I get the following warning when recovering:
```
Each child in a list should have a unique "key" prop.

Check the render method of `DisplayRecovery`
```

orhoj (approved):

PR #54, opened 2021-09-10
Add anonymity revoker update

shjortConcordium:
## Purpose

Adds the "Add Anonymity Revoker" chain update.
Companion to https://github.com/Concordium/concordium-ledger-app/pull/19.

## Changes

Adds to "Add Anonymity Revoker" update type, and the components to view/create/sign it.

## Checklist

      hard-to-understand areas.


app/features/ledger/SignAddIdentityProvider.ts
```diff
@@ -20,6 +21,36 @@ import {
 
 const INS_ADD_IDENTITY_PROVIDER = 0x2d;
 
+/**
+ * Sends the given description for signing.
+ * Uses the given p1 for length and p1 + 1 for the content.
+ */
+export async function sendDescription(
```
orhoj, line 28: This should be moved to a shared file instead of residing here, since it is used by two transactions and not just this one.

app/features/ledger/SignAddIdentityProvider.ts
```diff
@@ -20,6 +21,36 @@ import {
 
 const INS_ADD_IDENTITY_PROVIDER = 0x2d;
 
+/**
+ * Sends the given description for signing.
+ * Uses the given p1 for length and p1 + 1 for the content.
```
orhoj, line 26: This could be misunderstood as the `p1` containing some length. Adjust the documentation to make it more clear what is happening.

orhoj:
The `public key` field is not formatted for easy verification on the Ledger.

shjortConcordium:
> The `public key` field is not formatted for easy verification on the Ledger.

I noticed the same was the case for the VerifyKey hash/CDIVerifyKey in IdentityProvider, so I've also adjusted those.

orhoj (approved):

PR #63, opened 2021-09-13
Fix fetching transactions from wallet proxy failing silently

### LINKED ISSUE #41: Loading Transactions from wallet proxy fails silently
**Bug Description**

If the wallet is unable to reach the wallet proxy, when fetching new transactions,
there is no error given to the user.   

**Steps to Reproduce**
Have a local node running.
Disconnect from the internet.
Enter the account page.

**Expected Result**

After the request to the wallet proxy fails, an error should be shown.

**Actual Result**

Nothing happens.

**Versions**
- 1.1.6
- Linux

shjortConcordium:
## Purpose

Fix #41 .

## Changes

Reworked updateTransactions to return a promise, that gets resolved/rejected when the updating is done or fails.
AccountView now catches rejections from updateTransactions and displays an error to the user.

## Checklist

### DIFF
--- a/app/constants/errorMessages.json
+++ b/app/constants/errorMessages.json
@@ -1,5 +1,6 @@
 {
     "alreadyPrinting": "You seem to be printing already, please close your current print dialog before attempting to print again.",
     "missingGlobal": "Missing cryptographic parameters, please connect to a node to load these.",
+    "unableToReachWalletProxy": "Unable to reach the Wallet Proxy.",
     "unableToReachNode": "Unable to reach node."
 }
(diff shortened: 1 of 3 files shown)


app/constants/errorMessages.json
```diff
@@ -1,5 +1,6 @@
 {
     "alreadyPrinting": "You seem to be printing already, please close your current print dialog before attempting to print again.",
     "missingGlobal": "Missing cryptographic parameters, please connect to a node to load these.",
+    "unableToReachWalletProxy": "Unable to reach the Wallet Proxy.",
```
shjortConcordium, line 4: I guess we can't expect the user to know the wallet proxy, but what can we say instead?

orhoj: `An error occurred while trying to fetch new transactions.` or something like that.

orhoj (approved):

PR #64, opened 2021-09-13
Fix importing account can break convention 

### LINKED ISSUE #43: Importing account, when addressBookEntry already exists, breaks convention
**Bug Description**

Currently we have convention that for each account there should a corresponding entry on the addressbook, which has the same name ( and is readOnly).
This can be broken by having an addressBookEntry for with address A, and then importing the account. the already existing addressBookEntry will still exist and can be removed. (and the name can be different). 

**Steps to Reproduce**

Have an empty database and a backup with an account on address A.
Create an addressBookEntry on Address A.
import the backup.

**Expected Result**

After importing there should be a readOnly addressBookEntry on Address A, with the same name as the account. 

**Actual Result**

The addressBookEntry on address A is editable, and the name is the same as prior to the import.

**Versions**
- 1.1.6
- Ubuntu

shjortConcordium:
## Purpose

Fix #43 

## Changes

Import now updates `readOnly` status on existing `addressBookEntries`, when corresponding account is imported.
The `resolveAddressBookNameConflict` function now takes `readOnly` status into account.
Added possibility to add messages specifically to `addressBookEntries`, during import.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/utils/importHelpers.ts
+++ b/app/utils/importHelpers.ts
@@ -31,6 +31,7 @@ import {
 } from '~/features/AddressBookSlice';
 
 const alreadyExistsMessage = 'Already exists';
+const alreadyExistsAsAnAccount = 'Already exists as an account';
 const replacesMessage = (name: string) => `Replaces name: ${name}`;
 const alreadyExistsAsMessage = (name: string) => `Already exists as: ${name}`;
 const replacesPlaceholderMessage = (name: string) =>
@@ -50,7 +51,11 @@ interface AttachedEntities {
     attachedAccounts: Account[];
 }
 
-export type AddMessage = (id: string | number, message: string) => void;
+export type AddMessage = (
+    id: string | number,
+    message: string,
+    isAddressBookMessage?: boolean
+) => void;
 
 /**
  * Given two names of an account, and the account's address, determines the name to use.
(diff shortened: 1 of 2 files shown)


app/utils/importHelpers.ts
```diff
@@ -847,25 +881,25 @@ export async function importAddressBookEntries(
...
                     message,
-                } = await resolveAddressBookNameConflict(
-                    match.name,
-                    duplicate.name,
-                    address
-                );
+                } = await resolveAddressBookNameConflict(match, duplicate);
                 update.name = chosenName;
-                addMessage(address, message);
+                addMessage(address, message, true);
```
orhoj, line 895: Try to avoid parameterizing functions with a boolean. When you read
```
addMessage(address, message, true);
```
it is hard to know what `true` indicates. It becomes much simpler if it is just:
```
addAddressBookMessage(address, message);
```

shjortConcordium: I agree, I have refactored, let me know what you think.

app/utils/importHelpers.ts
```diff
@@ -50,7 +51,11 @@ interface AttachedEntities {
     attachedAccounts: Account[];
 }
 
-export type AddMessage = (id: string | number, message: string) => void;
+export type AddMessage<IdType = string | number> = (
+    id: IdType,
+    message: string,
+    isAddressBookMessage?: boolean
```
orhoj, line 57: You sould be able to remove this.

app/utils/importHelpers.ts
```diff
@@ -827,7 +861,7 @@ export async function importAddressBookEntries(
     dispatch: Dispatch,
     entries: AddressBookEntry[],
     addressBook: AddressBookEntry[],
-    addMessage: AddMessage
+    addMessage: AddMessage<string>
```
orhoj, line 864: Call this `addAddressBookMessage` instead.

orhoj (approved):

PR #66, opened 2021-09-13
Fix: Encrypted transfer fail if another is pending

### LINKED ISSUE #45: Encrypted transfer can fail if created while another is pending.
**Bug Description**

If an encrypted transfer is created while another is pending, but sent after the other has been finalized, then it will fail due to invalid proof. 

The same is likely the case for "unshield amount" transfers.

**Steps to Reproduce**

-  Send an encrypted transfer T1.
-  (Quickly) Construct another encrypted transfer and enter the submit page and begin signing T2 on ledger.
-  Wait until T1 has been finalized to complete signing T2 on ledger.

**Expected Result**

T2 is successful. (Or is not allowed to be sent)

**Actual Result**

T2 fails due to invalid proof. 

**Versions**
- 1.1.6
- Linux

shjortConcordium:
## Purpose

Fix #45.

## Changes

When the user tries to create a `shielded transfer` or a `unshield`, they get an error, if they have a pending `shield`/`unshield`/`shielded transfer`.
Now updates ledger component message in `SubmitTransfer.`
Refactored `FinalPage` out of Internal/External Transfer components, to simplify. (and avoid unnecessary calls to node)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/components/Transfers/ensureNoPendingShieldedBalance.tsx
+++ b/app/components/Transfers/ensureNoPendingShieldedBalance.tsx
@@ -16,10 +16,8 @@ export interface WithAccount {
 export default function ensureNoPendingShieldedBalance<
     TProps extends WithAccount
 >(Component: ComponentType<TProps>): ComponentType<TProps> {
-    const EnsureNoPendingShieldedBalanceComponent = ({
-        account,
-        ...props
-    }: TProps) => {
+    const EnsureNoPendingShieldedBalanceComponent = (props: TProps) => {
+        const { account } = props;
         const dispatch = useDispatch();
         const [showError, setShowError] = useState<boolean>(false);
 
@@ -37,7 +35,7 @@ export default function ensureNoPendingShieldedBalance<
                     header="Account has pending transfer impacting shielded balance"
                     onClick={() => dispatch(routerActions.goBack())}
                 />
-                <Component {...props} account={account} />
+                <Component {...props} />
             </>
         );
     };
--- a/app/components/Transfers/ensureNoPendingShieldedBalance.tsx
+++ b/app/components/Transfers/ensureNoPendingShieldedBalance.tsx
@@ -33,6 +33,7 @@ export default function ensureNoPendingShieldedBalance<
                 <SimpleErrorModal
                     show={showError}
                     header="Account has pending transfer impacting shielded balance"
+                    content="Please wait for any pending shield amounts, unshield amounts or shielded transfers to finalize."
                     onClick={() => dispatch(routerActions.goBack())}
                 />
                 <Component {...props} />
--- a/app/components/Transfers/ensureNoPendingShieldedBalance.tsx
+++ b/app/components/Transfers/ensureNoPendingShieldedBalance.tsx
@@ -1,0 +1,45 @@
+import React, { ComponentType, useState, useEffect } from 'react';
+import { useDispatch } from 'react-redux';
+import { routerActions } from 'connected-react-router';
+import { Account } from '~/utils/types';
+import { hasPendingShieldedBalanceTransfer } from '~/database/TransactionDao';
+import SimpleErrorModal from '~/components/SimpleErrorModal';
+
+export interface WithAccount {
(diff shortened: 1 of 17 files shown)


app/components/Transfers/ensureNoPendingShieldedBalance.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Account } from '~/utils/types';
+import { hasPendingShieldedBalanceTransfer } from '~/database/TransactionDao';
+import SimpleErrorModal from '~/components/SimpleErrorModal';
+
+export interface WithAccount {
+    account: Account;
+}
+
+/**
+ * Component that injects the next nonce on the given account.
```
orhoj, line 13: This is not what this does.

app/components/Transfers/ensureNoPendingShieldedBalance.tsx
```diff
@@ -0,0 +1,43 @@
...
+        }, [account.address]);
+
+        return (
+            <>
+                <SimpleErrorModal
+                    show={showError}
+                    header="Account has pending transfer impacting shielded balance"
+                    onClick={() => dispatch(routerActions.goBack())}
+                />
+                <Component {...props} />
```
orhoj, line 38: Should the component be hidden until we have completed the required check?

app/preload/database/transactionsDao.ts
```diff
@@ -80,6 +81,21 @@ async function hasEncryptedTransactions(
     return Boolean(transaction);
 }
 
+async function hasPendingShieldedBalanceTransfer(fromAddress: string) {
```
orhoj, line 84: You need to add some indices to support this query. Currently all it will use is the one on `fromAddress`, and for e.g. a baker account this will be insufficient to make this check effective. A reasonable assumption is that there will always be a limited amount of `Pending` transactions, so in this case you can probably just add an index for TransactionStatus. We will probably need an index on `transactionKind` at some point as well, so consider adding that now too.

app/pages/Accounts/SubmitTransfer/SubmitTransfer.tsx
```diff
@@ -195,14 +199,17 @@ export default function SubmitTransfer({ location }: Props) {
...
+            setMessage
         );
 
         const path = getAccountPath({
             identityIndex: credential.identityNumber,
             accountIndex: credential.credentialNumber,
             signatureIndex,
         });
+
+        setMessage('Please review and sign transaction on device ');
```
orhoj, line 212: Remove the extra space at the end.

app/components/Transfers/ensureNoPendingShieldedBalance.tsx
shjortConcordium: It is a local check, so it shouldn't be necessary.

app/preload/database/transactionsDao.ts
shjortConcordium: I'll add the indices.

app/components/Transfers/ensureNoPendingShieldedBalance.tsx
```diff
@@ -0,0 +1,43 @@
...
+                .then(setShowError)
+                .catch(() => {});
+            // eslint-disable-next-line react-hooks/exhaustive-deps
+        }, [account.address]);
+
+        return (
+            <>
+                <SimpleErrorModal
+                    show={showError}
+                    header="Account has pending transfer impacting shielded balance"
```
orhoj, line 35: Is it clear that a **shield** transaction is a transfer? Also, I think we should add some body text as it looks weird without. Perhaps it could say:
```
Please wait for any pending shielding transactions or encrypted transfers to finalize.
```

shjortConcordium: I've added it, but changed it to:

```
Please wait for any pending shield amounts, unshield amounts or shielded transfers to finalize.
```

orhoj (approved):

PR #69, opened 2021-09-15
Improve transaction insert efficiency

### LINKED ISSUE #67: Improve transaction insert efficiency
**Description**

Currently when inserting transactions, the app loads every transaction into memory to check if the transaction already exists.
This is very inefficient for large databases, and could potentially fail, if the transactions won't fit in the memory.

We should change this, to check for the duplicate transactions more efficiently and scalable.

shjortConcordium:
## Purpose

Fix #67.

## Changes

Stop loading all transactions when inserting transactions.
Rework AccountView effects so fetching transactions is not interrupted by accountInfo update.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/preload/database/transactionsDao.ts
+++ b/app/preload/database/transactionsDao.ts
@@ -4,7 +4,10 @@ import {
     TransactionStatus,
     TransferTransaction,
 } from '~/utils/types';
-import { transactionTable } from '~/constants/databaseNames.json';
+import {
+    transactionTable,
+    accountsTable,
+} from '~/constants/databaseNames.json';
 import { knex } from '~/database/knex';
 import { chunkArray, partition } from '~/utils/basicHelpers';
 import { GetTransactionsOutput } from '~/database/types';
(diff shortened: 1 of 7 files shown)


app/preload/database/transactionsDao.ts
```diff
@@ -89,7 +89,14 @@ export async function insertTransactions(
     transactions: Partial<TransferTransaction>[]
 ) {
     const table = (await knex())(transactionTable);
-    const existingTransactions: TransferTransaction[] = await table.select();
+
+    const hashes = transactions
+        .map((t) => t.transactionHash || '')
+        .filter((hash) => hash);
+    const existingTransactions: TransferTransaction[] = await table
+        .whereIn('transactionHash', hashes)
```
orhoj, line 97: `transactionHash` is not indexed, so this query will degrade over time.

app/preload/database/transactionsDao.ts
```diff
@@ -89,7 +89,14 @@ export async function insertTransactions(
     transactions: Partial<TransferTransaction>[]
 ) {
     const table = (await knex())(transactionTable);
-    const existingTransactions: TransferTransaction[] = await table.select();
+
+    const hashes = transactions
+        .map((t) => t.transactionHash || '')
```
orhoj, line 94: Can you flip the order of `map` and `filter` here? So it becomes:
```
.filter((t) => t.transactionHash).map(t.transactionHash);
```
That way we can perhaps get rid of the `|| ''`.

orhoj (approved):

PR #78, opened 2021-09-20
Load batch of newest transactions

### LINKED ISSUE #42: Fetch a batch of the latest non-reward transactions from the wallet proxy
**Task description**
Accounts that receive a lot of transactions, e.g. finalizers and bakers, have a problem finding non-reward transactions as they have to fetch all transactions from the wallet proxy before they can find their regular account transactions in between.

The main problem is that there can be thousands upon thousands of reward events that have to be processed first. To remedy this we will add an additional step when fetching from the wallet proxy, before performing the existing flow:

- Fetch the latest non-reward transactions from the wallet proxy.

This should result in the user always having their latest transactions available, which are assumed to be the most relevant. Afterwards we will continue to fetch old transactions in an ascending order as it is done today.

Notes:
- When the transaction fetching catches up we will end up fetching the same transactions again, and when that happens they can safely be ignored or be used to overwrite the existing entries. As long as they do not show up multiple times in the database everything should be fine.

shjortConcordium:
## Purpose

Have the wallet load a batch of the newest transactions with the current filters, so we user can see those, while loading older transactions.
Fix #42.

## Changes

Add function to request newest transactions from wallet-proxy.
Add effect to `useAccountSync`, which calls this function.
Added function to `transactionDao,` which is given a transactionId, and returns that specific transaction.

## Checklist

      hard-to-understand areas.


app/preload/database/transactionsDao.ts
```diff
@@ -34,6 +34,15 @@ async function getPendingTransactions(): Promise<TransferTransaction[]> {
     return transactions;
 }
 
+async function getTransaction(id: string): Promise<TransferTransaction> {
```
orhoj, line 37: This is typed incorrectly as it will return `undefined` if there are no results. You can add the typing by adding:
```
.table<TransferTransaction>(transactionTable)
```
as it will then return `TransferTransaction | undefined`. It can also be used `getTransactionsOfAccount`.

app/features/TransactionSlice.ts
```diff
@@ -277,6 +279,44 @@ export const reloadTransactions = createAsyncThunk(
...
+) {
+    const newestTransactionInDatabase = await getTransaction(
+        account.maxTransactionId
+    );
+
+    if (
+        account.transactionFilter.toDate &&
+        secondsSinceUnixEpoch(
+            new Date(account.transactionFilter.toDate)
+        ).toString() < newestTransactionInDatabase.blockTime
```
orhoj, line 299: It is a little dirty to do the comparison with strings. It would be preferable that `newestTransactionInDatabase.blockTime` was considered as a date/number instead.

app/pages/Accounts/useAccountSync.ts
```diff
@@ -35,9 +36,31 @@ export default function useAccountSync(): string | undefined {
     const dispatch = useDispatch();
     const account = useSelector(chosenAccountSelector);
     const accountInfo = useSelector(chosenAccountInfoSelector);
-    const [controller] = useState(new AbortController());
+    // This controller is used to abort updateTransactions, when the chosen account is changed, or the view is destroyed.
+    const [updateTransactionsController] = useState(new AbortController());
+    // This controller is used to control when the wallet should load a batch of the newest transactions.
+    const [newestTransactionsController] = useState(new AbortController(false));
```
orhoj, line 42: Try to do it without having a second controller, as we usually use controllers to make sure we don't have leaks, and not to control when something should start running.

shjortConcordium: I have removed the second controller, and added another field to the original controller.

orhoj (approved):

PR #80, opened 2021-09-20
Check signature is signed by authorized key.

### LINKED ISSUE #72: Wrong key is not rejected when attaching signatures
**Bug Description**
A wrong key could be attached to a governance proposal without showing an error. The proposal then showed an error when it was submitted to chain.

**Expected Result**
Signature is rejected already when proposer tries to add it to a proposal.

**Actual Result**
Signature was attached.

**Versions**
- DW 1.1.6

shjortConcordium:
## Purpose

For UpdateInstructions check that signatures are signed by authorized keys before adding to proposal. 
Fix #72.

## Changes

Changes to primary signing and proposal view for checking the signature is signed by an authorized key.
Make MultiSignatureCreateProposal use ensureChainData instead of withChainData.

## Checklist

- [ ] I have documented my code, in particular the intent of the
      hard-to-understand areas.


app/pages/multisig/ProposalView/util.ts
```diff
@@ -144,8 +150,31 @@ async function HandleUpdateInstructionSignatureFile(
...
+    let blockSummary;
+    try {
+        blockSummary = (await fetchLastFinalizedBlockSummary())
+            .lastFinalizedBlockSummary;
+    } catch {
+        return {
+            show: true,
+            header: errorMessages.unableToReachNode,
+            content:
+                'Unable to verify signature is signed by an authorized key without connection to node.',
```
orhoj, line 162: Unable to verify that the signature is signed by an authorization key without a connection to a node.

app/pages/multisig/ProposalView/util.ts
```diff
@@ -144,8 +150,31 @@ async function HandleUpdateInstructionSignatureFile(
...
+        };
+    }
+    const handler = findUpdateInstructionHandler(proposal.type);
+    try {
+        await attachKeyIndex(signature, blockSummary, proposal, handler);
```
orhoj, line 167: It is a little hacky to reuse `attachKeyIndex` for validation as you are not trying to attach the key index. Either create a new method with a `boolean` return type, or use `findKeyIndex()` directly and check that instead of `try/catch` on this method.

orhoj (approved):

PR #86, opened 2021-09-21
Fix constraint violations on insert

shjortConcordium:
## Purpose

Fix constraint violations, which are happening, when loading transactions (on finalizer accounts).

## Changes

Reworked insertTransactions and upsertTransactionsAndUpdateMaxId:
 - Check for transaction existence using transactionHash _and id_.
 - Changed insertTransactions to update transactionally like upsertTransactionsAndUpdateMaxId.
 - Refactored most of the behaviour into common delegate functions.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/preload/database/transactionsDao.ts
+++ b/app/preload/database/transactionsDao.ts
@@ -162,18 +162,19 @@ async function hasPendingShieldedBalanceTransfer(fromAddress: string) {
     return Boolean(transaction);
 }
 
-interface UpdatesAndAdditions<T> {
-    updates: T[];
-    additions: T[];
+interface UpdatesAndAdditions {
+    hashUpdates: TransferTransaction[];
+    idUpdates: TransferTransaction[];
+    additions: TransferTransaction[];
 }
 /**
  * Find which of the given transactions already exists in the database.
  * @returns updates, which are the transactions that already exist,
  * and additions, which are those that don't already exist.
  */
-async function findExistingTransactions<T extends Partial<TransferTransaction>>(
-    transactions: T[]
-): Promise<UpdatesAndAdditions<T>> {
+async function findExistingTransactions(
+    transactions: TransferTransaction[]
+): Promise<UpdatesAndAdditions> {
     const knexConnection = await knex();
     const hashes = transactions
         .map((t) => t.transactionHash || '')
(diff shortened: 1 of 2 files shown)


app/preload/database/transactionsDao.ts
```diff
@@ -161,45 +162,97 @@ async function hasPendingShieldedBalanceTransfer(fromAddress: string) {
...
+>(updates: T[], additions: T[], trx: Knex.Transaction): Promise<void> {
+    const additionChunks = chunkArray(additions, 50);
+
+    for (const additionChunk of additionChunks) {
+        await trx.table(transactionTable).insert(additionChunk);
+    }
+
+    for (const updatedTransaction of updates) {
+        const { transactionHash, ...otherFields } = updatedTransaction;
+        if (transactionHash) {
```
orhoj, line 226: If `findExistingTransactions` returned `{ hashUpdates, idUpdates, additions }`, then this condition could be removed.

app/preload/database/transactionsDao.ts
```diff
@@ -161,45 +162,97 @@ async function hasPendingShieldedBalanceTransfer(fromAddress: string) {
...
+ * */
+export async function insertTransactions(
+    transactions: Partial<TransferTransaction>[]
+) {
+    const { updates, additions } = await findExistingTransactions(transactions);
+    await (await knex())
+        .transaction(async (trx) => {
+            await upsertTransactionsTransactionally(updates, additions, trx);
+        })
+        .catch((e) => {
```
orhoj, line 253: To me it seems you're kind of mixing the styles suggested in the knex documentation. For the async/await case they enclose the transaction in a try/catch block instead of having this re-throw.

app/preload/database/transactionsDao.ts
shjortConcordium: Yes, I refactored it.

app/preload/database/transactionsDao.ts
shjortConcordium: I have removed this re-throwing. I added it because I, mistakenly, thought that knex was suppressing the error otherwise.

shjortConcordium:
I have changed insertTransaction's parameter to a TransferTransaction instead of a Partial one. This allowed me to remove the generic typing from the common functions.

orhoj (approved):

PR #87, opened 2021-09-21
Transactions loading/syncronizing indicator and account page more accessible without node connection

soerenbf:
## Purpose

Improve account page by adding indications of when transactions are loading/synchronizing, and making the account page work (in a limited sense) even when account has no `accountInfo` from node.

## Changes

- Ensure loading component takes priority over no transactions in list
- Added icon to show when account transactions are being synced with wallet proxy
- Made components on account page work with no account info
- Improved transaction loading logic (reducer)

## Checklist

      hard-to-understand areas.


app/features/TransactionSlice.ts
```diff
@@ -330,17 +242,19 @@ export const updateTransactions = createAsyncThunk<
         }
 
         async function updateSubroutine(maxId: bigint) {
-            if (!account || controller.isAborted) {
-                controller.finish();
```
shjortConcordium, line 334: Why have you removed the finish statements?

app/pages/Accounts/useAccountSync.ts
```diff
@@ -71,36 +95,43 @@ export default function useAccountSync(): string | undefined {
...
-            dispatch(updateTransactions({ controller, onError: setError }));
+            dispatch(
+                updateTransactions({
+                    controller,
+                    onError: setError,
+                })
+            );
         }
+
+        dispatch(updateTransactions({ controller, onError: setError }));
```
shjortConcordium, line 106: Why have you added this?

app/pages/Accounts/useAccountSync.ts
```diff
@@ -71,36 +95,43 @@ export default function useAccountSync(): string | undefined {
...
+
         // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [
-        account?.address,
-        accountInfo?.accountAmount,
-        account?.status,
-        controller.isAborted,
-    ]);
+    }, [account?.address, accountInfo?.accountAmount, account?.status]);
```
shjortConcordium, line 109: I might be misunderstanding what you doing, but the isAborted is in the deps, so the effect is re-run when the previous update has finished.

app/features/TransactionSlice.ts
```diff
@@ -1,4 +1,11 @@
-import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit';
+import {
+    createAsyncThunk,
+    createSlice,
+    PayloadAction,
+    isAnyOf,
+    isFulfilled,
+} from '@reduxjs/toolkit';
+import { Mutex, MutexInterface } from 'async-mutex';
```
shjortConcordium, line 8: You are using a new dependency, but you have not added it to package.json.

soerenbf: weird.. just ran `yarn add` as usual... I'll try again

app/features/TransactionSlice.ts
soerenbf: Is moved to `extraReducers` in slice in most recent update

app/features/TransactionSlice.ts
```diff
@@ -438,6 +341,172 @@ export const updateTransactions = createAsyncThunk<
...
+                    } else {
+                        state.transactions = action.payload.transactions;
+                    }
+                }
+
+                if (isLatest) {
+                    state.loadingTransactions = false;
+                }
+
+                controller?.finish();
```
shjortConcordium, line 417: The controller on updateTransactions should only finish, when we have finished updating transactions (or it has been aborted). 
Currently it seems to finish when we load the first batch of transactions.

app/features/TransactionSlice.ts
```diff
@@ -438,6 +341,172 @@ export const updateTransactions = createAsyncThunk<
...
+            }
+        );
+
+        builder.addMatcher(
+            isAnyOf(updateTransactions.rejected, updateTransactions.fulfilled),
+            (state, action) => {
+                state.synchronizing = false;
+                action.meta.arg.controller.finish();
```
shjortConcordium, line 425: The controller on updateTransactions should only finish, when we have finished updating transactions (or it has been aborted).
Currently it seems to finish when we load the first batch of transactions.

shjortConcordium: which also means it is not aborted, when you change the account / view

soerenbf: hmm.. doesn't happen for me. The reject/fulfill of the action should happen when the update is recursively done, which happens when I run it.

shjortConcordium: I'll show you tomorrow.

soerenbf: This, along with the sync issue we talked about earlier today, should be fixed now.

shjortConcordium:
If I have two accounts, and change the chosen account twice in quick succession, then the wallet will start spamming the wallet-proxy with request, even though it is caught up.

app/pages/Accounts/TransactionList/FiniteTransactionList.tsx
```diff
@@ -27,7 +30,7 @@ export default function FiniteTransactionList({
 
     if (showLoading) {
         return (
-            <div className="flex">
+            <div className={clsx('flex', styles.thickBlueSeparatorTop)}>
                 <Loading
```
shjortConcordium, line 34: This Loading's text should be capitalized.

shjortConcordium (commented):
If I change accounts multiple times, I sometimes end up in a state, where the transactions are never loaded, and the "loading transactions" are there permanently. (until I close the app)

soerenbf:
> If I change accounts multiple times, I sometimes end up in a state, where the transactions are never loaded, and the "loading transactions" are there permanently. (until I close the app)

This should be fixed now.

shjortConcordium (approved):

PR #110, opened 2021-09-29
Fix minor layout inconsistencies

### LINKED ISSUE #104: Minor layout inconsistencies
**Description**
1. 
There are inconsistencies when marking invalid user input. E.g. in multisig simple and scheduled transfers, an invalid memo only shows a red error message but the memo characters stay blue. The invalid amount turns red, not only a red error message.
coloring

In single sig transfers, the characters become read as well.
coloring_2

2.
Wrong spelling is indicated but DW does not seem to allow the usual right-click context menu, which would allow auto correction.
spelling

**Versions**
- dw 1.2.0 beta2
- Windows
- Chrome

shjortConcordium:
## Purpose

Fix #104

## Changes

Make memo text error-colored on error in multisig flow.
Add context menu that allows cut/copy/paste and spellchecking.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/main/menu.ts
+++ b/app/main/menu.ts
@@ -5,6 +5,7 @@ import {
     shell,
     dialog,
     app,
+    MenuItem,
 } from 'electron';
 import { licenseNotices, supportForum } from '~/constants/urls.json';
 import { openRoute } from '~/constants/ipcRendererCommands.json';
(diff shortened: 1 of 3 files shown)


app/main/menu.ts
```diff
@@ -81,3 +82,49 @@ export function createMenu(window: BrowserWindow) {
...
+                enabled: params.editFlags.canPaste,
+            },
+        ]);
+
+        if (params.dictionarySuggestions.length > 0) {
+            menu.append(new MenuItem({ type: 'separator' }));
+        }
+
+        // Add each spelling suggestion
+        for (const suggestion of params.dictionarySuggestions) {
```
soerenbf, line 103: How many are suggestions are included? Could it be possible that the list could be too long, making the context menu huge, or is this limited by electron or on OS level?

shjortConcordium: We should check it on each OS, but this is how electron does it in their own documentation:
https://www.electronjs.org/docs/tutorial/spellchecker

shjortConcordium: On linux it does seem to never exceed 5 suggestions.

soerenbf: I haven't had it exceed 5 either on macOS (tried 3 times), maybe it's fine then 🙂

soerenbf (approved):

PR #119, opened 2021-10-01
fix validation based on current date

### LINKED ISSUE #116: Date and time errors when making account reports
**Bug Description**
When making an account report and setting the "To:" time too close to the current time, yields and error.

The wallet says "Date must be before today", even though it should be possible to make the report from today.

If the time is set to an hour before the current time, the wallet doesn't give you any errors.

Image 30-09-2021 at 15 36
**Steps to Reproduce**

**Expected Result**
It should be possible to make a report all the way up to the current time. If the user enters a time in the future, the error message should not be "Date must be before today", but something like "The time cannot be in the future" or something.

**Actual Result**
It is not possible to make a report that goes all the way up past the last hour, and it says "Date must be before today"

**Versions**
- Software Version: 1.2.0-beta.2
- OS: MacOS 11.5.2

soerenbf:
## Purpose

Fix #116

### DIFF
--- a/app/components/TransactionFilters.tsx
+++ b/app/components/TransactionFilters.tsx
@@ -140,7 +140,9 @@ const getGroupValues = (group: TransactionKindString[], v: boolean) =>
         {} as Partial<TransactionFilter>
     );
 
-const pastDateValidator = allowOptional(pastDate('The time cannot be in the future'));
+const pastDateValidator = allowOptional(
+    pastDate('The time cannot be in the future')
+);
 
 type Callback = (filter: TransactionFilter) => Promise<unknown>;
 
--- a/app/components/TransactionFilters.tsx
+++ b/app/components/TransactionFilters.tsx
@@ -140,7 +140,7 @@ const getGroupValues = (group: TransactionKindString[], v: boolean) =>
         {} as Partial<TransactionFilter>
     );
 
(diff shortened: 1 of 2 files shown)


shjortConcordium:
Something else that Jens mentioned in the issue was that the error message you get is "Date must be before today" because the rule is actually that the time cannot be in the future / must be before now.
Can you also fix that here?

soerenbf:
> Something else, that Jens mentioned in the issue, was that the error message you get is "Date must be before today", but the rule is actually that the time cannot be in the future / must be before now. Can you also fix that here?

Yes, completely forgot.. thanks for the reminder! 😄

app/components/TransactionFilters.tsx
```diff
@@ -140,7 +140,7 @@ const getGroupValues = (group: TransactionKindString[], v: boolean) =>
         {} as Partial<TransactionFilter>
     );
 
-const pastDateValidator = allowOptional(pastDate('Date must be before today'));
+const pastDateValidator = allowOptional(pastDate('Cannot be future date'));
```
shjortConcordium, line 143:
```suggestion
const pastDateValidator = allowOptional(pastDate('The time cannot be in the future'));
```
instead maybe?

shjortConcordium (approved):

PR #122, opened 2021-10-01
Improvements to Account Report

shjortConcordium:
## Purpose

Currently, when creating accounts reports, we load everything into memory (max: 1M transactions) during the process. This leads to problems when creating big/multiple accounts reports.

## Changes

Change account report to "stream" transactions into the output file/archive.
Most of the process is moved to the preload script. 
Display a modal while creating the report (Which allows the user to abort the process).
Moving around some helper functions.
 
## Checklist

      hard-to-understand areas.


shjortConcordium:
The "error" where not all transactions got into the report, was more an error of measurement, between each batch (10000) transactions, there was not a line break. This caused it to seem some transactions were missing, but they were just sharing lines.

shjortConcordium:
Removed the tests for `toCSV`. I moved the function into `preload/accountReport.ts`, and simplified it / removed explicitly putting in the headers in the top.

app/utils/transactionHelpers.ts
```diff
@@ -705,3 +688,36 @@ export function isOutgoingTransaction(
...
+        ].includes(transaction.transactionKind) &&
+        transaction.fromAddress !== currentAddress
+    );
+}
+
+/**
+ * Determine whether the transaction affects shielded balance.
+ */
+export function isShieldedBalanceTransaction(
+    transaction: Partial<TransferTransaction>
```
orhoj, line 712: I know you just moved this method, but `Partial<TransferTransaction>` allows for `transaction.transactionKind` to not be set. Perhaps you can improve it so that we do not allow it to be optional here.

app/preload/database/addressBookDao.ts
```diff
@@ -47,6 +47,18 @@ async function findEntries(
         .then((e) => e.map(sanitizeAddressBookEntry));
 }
 
+export async function getEntryName(address: string) {
+    const entry = await (await knex())
+        .select<AddressBookEntry>()
```
orhoj, line 52: You should be able to select the name directly in the SQL query:
```
.select('name')
```

orhoj, line 52: You should be able to select the name directly in the SQL query:
```
.select('name')
```

app/utils/addressBookHelpers.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export async function lookupAddressBookEntry(
+    address: string
+): Promise<AddressBookEntry | undefined> {
+    const entries = await findEntries({ address });
+    return entries[0];
+}
+
+/**
+ * Attempts to find the address in the accounts, and then AddressBookEntries
```
orhoj, line 12: As far as I can see this only checks something in the address book table, and does not check anything in the accounts table.

app/pages/Accounts/AccountReport/AccountReport.tsx
```diff
@@ -149,6 +163,13 @@ export default function AccountReport({ location }: Props) {
                 content={showError.content}
                 onClick={() => setShowError({ show: false })}
             />
+            <SimpleErrorModal
```
orhoj, line 166: It's quite hacky to use one of our error modals for an info message. We should use another component for this.

app/preload/accountReport.ts
```diff
@@ -0,0 +1,274 @@
...
+    abortController: AbortController
+) {
+    const getAddressName = async (address: string) => {
+        if (!address) {
+            return undefined;
+        }
+        if (address === account.address) {
+            return account.name;
+        }
+        return getEntryName(address);
```
orhoj, line 141: A potential optimization (that we probably don't need right now) would be to cache the names we look up here, as if we see an address once there is probably a high chance of seeing it again.

app/utils/transactionHelpers.ts
shjortConcordium: We don't actually need it to be partial, it turns out.

app/preload/database/addressBookDao.ts
shjortConcordium: Yes.

app/utils/addressBookHelpers.ts
shjortConcordium: Yes, this comment was way outdated.

app/pages/Accounts/AccountReport/AccountReport.tsx
shjortConcordium: I've used the `MessageModal` instead

app/preload/accountReport.ts
shjortConcordium: Yea, I did the obvious one, with the account, which are making the report.

orhoj:
The report seems to be missing some transactions when I compare it to a count against the transaction table.

orhoj:
The chosen filters are not always applied to the transaction report. If I exit the view and start over, then they work, but if I make multiple exports, then I run into the problem.

PR #131, opened 2021-10-05
Change name to Concordium Desktop Wallet

### LINKED ISSUE #126: Streamline naming: Concordium Desktop Wallet
**Description**
At the moment we are not completely consistent in how we use the names _Concordium Wallet_ and _Concordium Desktop Wallet_. It would be good if we could just use the name _Concordium Desktop Wallet,_ to also keep it consistent with how we name the mobile wallet. 

As there could be some complexity related to how paths are named, it might not be straight forward to rename the desktop wallet.

**Versions**
- Software Version: 1.2.0-beta.2
- OS: MacOS, Linux, Windows

shjortConcordium:
## Purpose

Closes #126 

## Changes

Changes the productName to "Concordium Desktop Wallet" (+ TARGET_NET).
Added a userData variable, and make the app use that to determine the userData folder. 

## Checklist

      hard-to-understand areas.


configs/partials/webpack.config.base.js
```diff
@@ -11,12 +11,16 @@ const { dependencies: externals } = require('../../app/package.json');
 
 const extensions = ['.js', '.jsx', '.json', '.ts', '.tsx'];
 
-if (
-    process.env.TARGET_NET &&
-    !['stagenet', 'testnet'].includes(process.env.TARGET_NET)
-) {
+const targetNet = process.env.TARGET_NET;
+
+let userData = process.env.npm_package_userData;
```
orhoj, line 17: I don't see the benefit of having `userData` defined in `package.json`. I think we might as well just set it here.

shjortConcordium: I've moved it.
I was probably thinking it would be easier to find there, but it should never have to be changed anyway. :shrug:

orhoj: Well... If we could migrate it easily, then we should do that now. And otherwise it is quite likely remaining quite static.

configs/partials/webpack.config.base.js
```diff
@@ -12,6 +12,19 @@ const CheckTargetNet = require('../../internals/scripts/CheckTargetNet');
...
 CheckTargetNet();
 
+const targetNet = process.env.TARGET_NET;
+
+let userData = 'Concordium Wallet';
+if (targetNet) {
+    userData += ` ${targetNet}`;
+}
+
+if (targetNet && !['stagenet', 'testnet'].includes(targetNet)) {
```
orhoj, line 22: Why was this check added? We already have `CheckTargetNet()` above.

shjortConcordium: Some sloppy merging from my side.
Though it did make me notice another error. :)

orhoj (approved):

PR #144, opened 2021-10-18
Use node sdk

shjortConcordium:
## Purpose

Use https://github.com/Concordium/concordium-node-sdk-js where possible for grpc calls.

## Changes

Now uses the node-sdk for all grpc requests, expect `SendTransaction`

Changed types to use/fit the types from the node-sdk.
Revamped the preload interface for grpc, removing the `performGrpcCall` and adding endpoints for the individual calls.
Moved `getStatus` from `transactionHelpers` to `nodeHelpers`.
Added `toStringBigInts`, in order to make sure we don't insert bigints into the database (from AccountInfo) 

## Checklist

      hard-to-understand areas.


app/node/nodeHelpers.ts
```diff
@@ -118,3 +126,45 @@ export async function nodeSupportsMemo() {
...
+        const interval = setInterval(async () => {
+            let response;
+            try {
+                response = await getTransactionStatus(transactionHash);
+            } catch (err) {
+                // This happens if the node cannot be reached. Just wait for the next
+                // interval and try again.
+                return;
+            }
+            if (!response) {
```
orhoj, line 155: It would be nice with a comment above this line stating that this is the case if the transaction is absent.

app/node/nodeHelpers.ts
```diff
@@ -118,3 +126,45 @@ export async function nodeSupportsMemo() {
...
+                clearInterval(interval);
+                resolve({ status: TransactionStatus.Rejected, outcomes: {} });
+                return;
+            }
+
+            if (response.status === 'finalized') {
```
orhoj, line 161: I believe this check should be:
```
response.status === TransactionStatusEnum.Finalized
```

app/pages/Accounts/AccountDetailsPage/CredentialInformation/CredentialInformation.tsx
```diff
@@ -52,20 +42,21 @@ export default function CredentialInformation({ account, accountInfo }: Props) {
...
 
     const credentials = Object.values(accountInfo.accountCredentials)
-        .map((o) => o.value.contents)
+        .map((o) => o.value)
         .map((cred) => {
-            let enrichedCred: CredentialOfAccount = {
-                ...cred,
+            const enrichedCred = {
+                ...cred.contents,
+                // The node returns the credId in the regId field for
```
orhoj, line 49: Since we now have a check on the type, the comment is no longer necessary to me. It is no longer a hack, as we actually check instead of just take whatever field was set.

app/pages/Accounts/AccountDetailsPage/CredentialInformation/DisplayIdentityAttributes.tsx
```diff
@@ -10,22 +11,21 @@ import {
 const attributeNames: Record<string, string> = attributeNamesJson;
 
 interface Props {
-    credential: CredentialDeploymentValues;
+    revealedAttributes: Record<AttributeKey, string>;
 }
 
 /**
  *  DIsplays the revealed Attributes of the chosen account.
```
orhoj, line 18: Update the method documentation here.

app/pages/Accounts/AccountDetailsPage/CredentialInformation/DisplayIdentityAttributes.tsx
```diff
@@ -10,22 +11,21 @@ import {
...
+    revealedAttributes,
 }: Props): JSX.Element | null {
-    const attributes = credential.policy.revealedAttributes;
-    const attributeKeys = Object.keys(attributes);
+    const attributeKeys = Object.keys(revealedAttributes);
 
     if (attributeKeys.length === 0) {
         return (
-            <div className="pT10" key={credential.credId || credential.regId}>
+            <div className="pT10">
```
orhoj, line 28: I'm guessing you will now get an error if you display 2 credentials on an account, where neither has a revealed credential. So you probably need to add some other key.

app/pages/Recovery/util.ts
```diff
@@ -103,7 +103,9 @@ function extractCredentialIndexAndPolicy(
         accountInfo.accountCredentials
     ).find(
         ([, cred]) =>
-            (cred.value.contents.credId || cred.value.contents.regId) === credId
+            (cred.value.type === 'initial'
```
orhoj, line 106: We do this quite a bit. Perhaps a helper function taking a credential and returning the `credId` would be nice to have.

app/pages/multisig/AccountTransactions/UpdateAccountCredentials/UpdateCredentialsPage.tsx
```diff
@@ -301,14 +308,18 @@ function UpdateCredentialPage({ exchangeRate }: Props): JSX.Element {
             currentAccountInfo.accountCredentials
         ).map((accountCredential) => {
             const credentialIndex = parseInt(accountCredential[0], 10);
-            const cred = accountCredential[1].value.contents;
-            if (cred.regId) {
+            const cred = accountCredential[1].value;
+            // TODO: don't use literal string
```
orhoj, line 312: Why not?

app/utils/JSONHelper.ts
```diff
@@ -40,3 +40,23 @@ export function intToString(jsonStruct: string, key: string) {
         `"${key}":"$1"`
     );
 }
+
+// Given any object/array, turn all bigint fields to strings;
```
orhoj, line 44: Use
```
/**  */
```
when documenting functions.

orhoj:
The node status in the lower left corner shows that the node is unavailable, even if I have a working connection to one.

app/components/ScheduleList/ScheduleList.tsx
```diff
@@ -1,14 +1,14 @@
 import React from 'react';
```
orhoj, line 1: If I try to inspect a release schedule, then I get a gray screen.

app/pages/multisig/AccountTransactions/UpdateAccountCredentials/UpdateCredentialsPage.tsx
shjortConcordium: I was thinking it would be nicer if there was an enum, probably.

app/pages/Accounts/AccountDetailsPage/CredentialInformation/DisplayIdentityAttributes.tsx
shjortConcordium: This shouldn't have a key, because it is not the responsibility of this component to provide a key, but the "list"-component. 
And this is used is not used directly in a list, so we shouldn't get an error.

app/utils/JSONHelper.ts
shjortConcordium: removed the function as it didn't work correctly.

shjortConcordium:
> The node status in the lower left corner shows that the node is unavailable, even if I have a working connection to one.

Should work now, it was failing silently

app/utils/credentialHelper.ts
```diff
@@ -96,3 +101,17 @@ export function getNoteForOwnCredential(
 
     return `Credential from "${identityName}"`;
 }
+
+/**
+ * Given a versioned credential, reutrns the credId.
```
orhoj, line 106: - reutrns -> returns
- Unsupprted -> Unsupported

app/pages/multisig/AccountTransactions/UpdateAccountCredentials/UpdateCredentialsPage.tsx
orhoj: But that would be the responsibility of the SDK to provide that. We can consider adding that later, but I don't think the TODO should be in the desktop wallet.

app/components/ScheduleList/ScheduleList.tsx
orhoj: This still happens to me.

app/pages/multisig/AccountTransactions/UpdateAccountCredentials/UpdateCredentialsPage.tsx
shjortConcordium: I've removed it.

app/utils/credentialHelper.ts
shjortConcordium: I are a amazing speller.

app/components/ScheduleList/ScheduleList.tsx
shjortConcordium: I fixed it and, and removed the "bigint in state" errors.

app/utils/JSONHelper.ts
```diff
@@ -1,23 +1,35 @@
...
+    Date: 'date',
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function replacer(this: any, k: string, v: any) {
+    if (typeof v === types.BigInt) {
+        return { '@type': types.BigInt, value: v.toString() };
+    }
+    if (this[k] instanceof Date) {
+        return { '@type': types.Date, value: v };
```
orhoj, line 13: Why do we need special handling of Dates now?

shjortConcordium: We need the special handling of Dates, because AccountInfo now contain dates (instead of just strings).
We don't have dates in the "wallet" export, and when exporting a scheduled transfer, the date is a timestamp string.

orhoj (approved):

Issue #33, opened 2021-08-25
Add single signer baker flows


jens-concordium:
**Description**
At the moment baker transactions are only available via the multi-sig flows, so it would be nice to have flows for single signers. 

A suggestion for the single signer flows can be found in Figma, on the page "Single signer baker transactions". This solution is dependent on "Advanced Account View" mentioned in issue #24.

This is a nice-to-have CR, as the functionality already exists.

soerenbf:
Agreed with @jens-concordium to deviate from the design, instead following the current convention when it comes to signing in other single-signer flows. This means having full-screen focus on the transaction being signed, along with the component showing the signing process.

PR #150, opened 2021-11-02
Allow user to verify address

shjortConcordium:
## Purpose

Allow user to verify their address on device.
Companion to https://github.com/Concordium/concordium-ledger-app/pull/33.

## Changes

Added a button to showAccountAddress to allow verifying address on ledger.

## Checklist

      hard-to-understand areas.


app/pages/Accounts/AccountDetailsPage/AccountDetailsPage.tsx
```diff
@@ -76,7 +76,11 @@ export default withAccountSync(function DetailsPage() {
                         component={accountChanged ? ToAccounts : BuildSchedule}
                     />
                     <Route path={routes.ACCOUNTS_ADDRESS}>
-                        <ShowAccountAddress account={account} asCard />
+                        <ShowAccountAddress
```
orhoj, line 79: You have only added the verification here, but not in the `TransactionsAndAddress` view.

orhoj:
If I try to verify the address of an account for an identity with `version = 0`, then the address shown on the Ledger does not correspond to the address in the wallet. The address on the Ledger is the one that is incorrect.

orhoj:
In Ledger live they don't show the address until it is also shown on the Ledger. Should we do something similar? (For the accounts where it is possible to verify the address).

orhoj:
The Ledger component stays open after having verified an account address. I think it would be nice if it was closed down again afterwards.

app/features/ledger/ExportPrivateKeySeed.ts
```diff
@@ -20,14 +31,15 @@ function requestKeys(
     return transport.send(0xe0, INS_EXPORT_PRIVATE_KEY_SEED, p1, p2, data);
 }
 
-export async function getPrivateKeySeeds(
+export async function getPrivateKeys(
```
orhoj, line 34: The updated methods could use some documentation stating what the `version` variable does. Because if the method is called `getPrivateKeys`, then it is not obvious that if `version === 1` that it returns seeds and not the final keys.

app/pages/Accounts/VerifyAddress.tsx
```diff
@@ -0,0 +1,104 @@
...
+            );
+        }
+
+        const credential = await findLocalDeployedCredentialWithWallet(
+            account.address,
+            ledger
+        );
+        if (credential === undefined) {
+            throw new Error(
+                'Unable to decrypt shielded balance and encrypted transfers. Please verify that the connected wallet is for this account.'
```
orhoj, line 39: Fix the error message.

app/pages/Accounts/VerifyAddress.tsx
```diff
@@ -0,0 +1,104 @@
...
+                credential.identityNumber,
+                credential.credentialNumber
+            );
+        }
+    }
+
+    return (
+        <>
+            <MessageModal
+                title="Unable to verify Address."
```
orhoj, line 56:
```suggestion
                title="Unable to verify account address."
```

app/pages/Accounts/VerifyAddress.tsx
```diff
@@ -0,0 +1,104 @@
...
+                message="An account's address is computed using the first credential, and your current Ledger device did not generate that."
```
orhoj, line 57:
```suggestion
                message="An account's address is derived from the initial credential on that account. The initial credential on this account was not created from the currently connected Ledger, and therefore it is not possible to verify the address of the account.
```

app/pages/IdentityIssuance/PickProvider/PickProvider.tsx
```diff
@@ -220,6 +220,7 @@ export default function IdentityIssuanceChooseProvider({
                     ledgerCallback={withLedger}
                     preCallback={getIdentityNumber}
                     disabled={!provider}
+                    identityVersion={0}
```
orhoj, line 223: I feel like I have seen this before. Hardcoding the version seems like a bad idea.

orhoj (changes requested):
I have added my review comments.

app/pages/Accounts/AccountDetailsPage/AccountDetailsPage.tsx
shjortConcordium: Yes, but that was deliberately, due to concerns regarding vertical space, 
but i can add it to the TransactionsAndAddress view, if you think it should be there

app/pages/IdentityIssuance/PickProvider/PickProvider.tsx
shjortConcordium: I have explained the hardcoding now.

app/pages/Accounts/AccountDetailsPage/AccountDetailsPage.tsx
orhoj: Get some input from Jens on how to design it. It should be an additional step to get your account address without verifying it at the same time, if we want to help users protect themselves.

app/pages/IdentityIssuance/PickProvider/PickProvider.tsx
orhoj: I think the better way to do this would be to change the name of the parameter. If the parameter was an enum then you could have it as:
```
Seed = 0,
Key = 1
```
and that way you would supply it like:
```
exportType={Seed}
```
making it very clear what is happening. It is a little hacky to call the input `identityVersion` when in reality the function doesn't care about the identity, it just has to know whether to get the seed or the key.

app/pages/Accounts/VerifyAddress.tsx
```diff
@@ -1,65 +1,81 @@
...
+    const reasonForNotBeingVerifiable = useAsyncMemo(
+        async () => {
+            if (identity === undefined) {
+                return 'No identity was found for this account. This is an internal error that should be reported.';
+            }
+            if (!(await hasFirstCredential(account.address))) {
+                return "An account's address is derived from the initial credential on that account. The initial credential is not in this wallet, and therefore it is not possible to verify the address of the account.";
+            }
+            if (identity.version === 0) {
+                return 'The identity used to create this account uses a deprecated version of the key generation. It is recommended to create a new identity and use accounts on that instead.';
```
orhoj, line 49: I am not certain that we should recommend people to do this.

app/utils/identityHelpers.ts
```diff
@@ -156,3 +157,10 @@ export function isRejectedIdentity(
         identity.status === IdentityStatus.RejectedAndWarned
     );
 }
+
+export function getKeyExportType(version: IdentityVersion) {
```
orhoj, line 161: Please document this function.

app/pages/Accounts/VerifyAddress.tsx
```diff
@@ -0,0 +1,133 @@
...
+                            <>
+                                {opened || (
+                                    <div
+                                        className={clsx(
+                                            className,
+                                            'flexColumn'
+                                        )}
+                                    >
+                                        <h3 className="mB40 mT0">
+                                            Unlock Ledger to verify and show
```
orhoj, line 111: This text is not centered. Also I am not sure about the text. It asks me to unlock even though my device is unlocked and connected. Perhaps we don't even need the text? Perhaps Jens can help us here...

orhoj (changes requested):
I need to be in the office to test further with a Ledger device. But I have found a few things and added comments for those already.

app/pages/Accounts/VerifyAddress.tsx
```diff
@@ -0,0 +1,133 @@
...
+                                            address
+                                        </h3>
+                                        {statusView}
+                                        <Button
+                                            size="big"
+                                            disabled={!isReady}
+                                            className="m40"
+                                            onClick={submitHandler}
+                                        >
+                                            Show and Verify Address
```
orhoj, line 121:
```suggestion
                                            Show and verify address
```

app/pages/Accounts/VerifyAddress.tsx
shjortConcordium: Removed it.

shjortConcordium:
> I don't get an error if I try to verify the address against an emulator.

Yes, the error was not handled properly, should be fixed now. (Same problem happening when using an app without verifyAddress)

app/pages/Accounts/VerifyAddress.tsx
shjortConcordium: Changed it to:
```
The address cannot be verified, 
because the identity used to create this account uses a deprecated version of the key generation, 
which is not available on the ledger.
```

app/pages/Accounts/VerifyAddress.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const reasonForNotBeingVerifiable = useAsyncMemo(
+        async () => {
+            if (identity === undefined) {
+                return 'No identity was found for this account. This is an internal error that should be reported.';
+            }
+            if (!(await hasFirstCredential(account.address))) {
+                return "An account's address is derived from the initial credential on that account. The initial credential is not in this wallet, and therefore it is not possible to verify the address of the account.";
+            }
+            if (identity.version === 0) {
+                return 'The address cannot be verified, because the identity used to create this account uses a deprecated version of the key generation, which is not available on the ledger.';
```
orhoj, line 49:
```suggestion
                return 'The address cannot be verified, because the identity used to create this account uses a deprecated version of the key generation, which is not available on the Ledger.';
```

orhoj:
When I finish viewing the address on the Ledger and press both buttons, then the address disappears from the UI in the desktop wallet. It should not disappear like that because I would want to copy the address afterwards.

orhoj:
It is inconsistent that we show the address on the address button, but you have to verify it on the Ledger before it is shown on the right side view. Can you talk to @jens-concordium about how this could be nicer?

jens-concordium:
We can just remove the address from the address button. The current "Change view" button will become the address button in the coming design update, so we'll get rid of the address directly on the button by then anyway.

orhoj:
> In Ledger live they don't show the address until it is also shown on the Ledger. Should we do something similar? (For the accounts where it is possible to verify the address).

I know I wrote this, but Ledger Live also allows you to state that you don't have your device, and then show the address anyway. Consider if we need the same. I think their UI flow is quite nice.

shjortConcordium:
> When I finish viewing the address on the Ledger and press both buttons, then the address disappears from the UI in the desktop wallet. It should not disappear like that because I would want to copy the address afterwards.

Well, I did it the same as in Ledger Live, when you have verified the address, it is hidden again. (I assume you should just copy it while verifying it)

orhoj (approved):

PR #151, opened 2021-11-03
Datepicker

### LINKED ISSUE #124: Change datetime inputs to datepicker component
**Description**
To improve usability and accessibility, we have decided to change datetime inputs to use a datepicker instead.

We have landed on the following library: [react-datepicker](https://www.npmjs.com/package/react-datepicker), as it is simple, features date + time, and is accessible + customizable.

As design options might be limited, implementation will probably end up deciding design, so collaboration with @jens-concordium is important.

soerenbf:
## Purpose

Replaces timestamp input with proper datepicker.

Closes #124 

## Changes
- Deleted timestamp input
- Added datepicker component
- Small layout adaptions where needed to make UI behave with new component

## Checklist

      hard-to-understand areas.


app/components/Form/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,85 @@
...
+            {label && <Label className="mB5">{label}</Label>}
+            <ReactDatePicker
+                selected={value}
+                onChange={onChange}
+                onBlur={onBlur}
+                showTimeSelect
+                timeCaption="Time"
+                timeFormat="HH:mm"
+                timeIntervals={15}
+                dateFormat="dd-MM-yyyy 'at' HH:mm:ss"
```
shjortConcordium, line 71: When we display dates, we usually use `yyyy-MM-dd`, maybe it should also be like that here?

soerenbf: I agree, they should definitely be aligned.. but doesn't it make more sense to display it with year last, as dates are normally displayed?

app/utils/timeHelpers.ts
```diff
@@ -1,3 +1,4 @@
+import { setHours, setMinutes, setSeconds, setMilliseconds } from 'date-fns';
```
shjortConcordium, line 1: Can we just use functions from a transitive dependency? :thinking:

app/utils/timeHelpers.ts
```diff
@@ -241,3 +242,20 @@ export function subtractHours(hours: number, date: Date) {
 export function parseTime(timeStamp: string | bigint, unit?: TimeStampUnit) {
     return getFormattedDateString(dateFromTimeStamp(timeStamp, unit));
 }
+
+const stripTime = (date: Date) =>
+    setHours(setMinutes(setSeconds(setMilliseconds(date, 0), 0), 0), 0);
```
shjortConcordium, line 247: wouldn't it be easier to just clone and clone.setHours(0,0,0,0)?

app/components/Form/DatePicker/DatePicker.tsx
shjortConcordium: Is ISO 8601 not the norm?

shjortConcordium: is ISO 8601 not the norm?

app/utils/timeHelpers.ts
soerenbf: You can, but it's obviously pretty bad to do.. I'll make sure they're installed.

app/components/Form/DatePicker/DatePicker.tsx
soerenbf: seems like it is 😅 I'll adapt.

shjortConcordium (approved):

PR #157, opened 2021-11-09
fix wrongly colored icon.

### LINKED ISSUE #152: Wrong icon color
**Bug Description**
The `(!)` icon on e.g. a failed identity is black instead of the previous red color that it had.

**Versions**
- Software Version 1.2.0

soerenbf:
## Purpose

Fix #152 

## Changes
- Ensure icon color is only overridden when no default color is already set.

## Checklist

      hard-to-understand areas.


app/styles/elements/_icon.scss
```diff
@@ -1,3 +1,3 @@
-svg {
+svg:not([fill]) {
```
shjortConcordium, line 1: This now overrides some svg's natural colors, it seems. Check the rename button, copy buttons, and icons in the addressBook view.
Screenshot from 2021-11-11 15-28-01
Screenshot from 2021-11-11 15-27-21

soerenbf: Good catch, will fix!

shjortConcordium (approved):

PR #158, opened 2021-11-11
Fix usage of getAccountInfo with credId's.

shjortConcordium:
## Purpose

After using nodejs sdk, calling getAccountInfo with a credId failed.
This should accommodate the fix made in the sdk (https://github.com/Concordium/concordium-node-sdk-js/pull/19). 

## Changes

Add a method for getting accountInfo with a credId.
Use that method in account recovery and account slice (for genesis accounts)
Refactoring in account slice to handle genesis accounts and confirmed accounts separately.

## Checklist

      hard-to-understand areas.


app/node/nodeHelpers.ts
```diff
@@ -15,46 +15,26 @@ import {
...
 
 /** Gets the accountInfo for the given address. */
 export async function getAccountInfoOfAddress(
     address: string
 ): Promise<AccountInfo> {
     const blockHash = await getlastFinalizedBlockHash();
-    return getAccountInfo(address, blockHash);
+    const accountInfo = await getAccountInfo(address, blockHash);
+    if (!accountInfo) {
+        throw new Error(`Address:${address} does not represent an account.`);
```
orhoj, line 35: No `accountInfo` can also be because the node is not caught up, so I don't think the error message is correct.

shjortConcordium: I have changed the error message.

PR #159, opened 2021-11-16
Stop storing transactions from the wallet proxy

orhoj:
## Purpose
To stop storing transactions loaded from the wallet proxy locally. The main reason for this is that we had to rely on the `id` field from the wallet proxy, which may change over time. This could result in missing transactions and/or duplicate transactions shown in the UI.

## Changes
- Stopped storing transactions received from the wallet proxy locally.
- Always get transactions from the wallet proxy when loading more transactions.
- Dropped the `maxTransactionId` column in the accounts table, as it is now unused.
- Fixed bug when decrypting outgoing encrypted transfers where the result was incorrect.
- Improved decryption performance by using pre-calculated table (like mobile wallets).
- Fixed an issue where the infinite scrolling list did not allow the user to scroll, which meant that not all transactions could be shown.
- Fixed a UI issue in the infinite scrolling list of transactions, where a newly received transaction would offset the UI incorrectly.

## Checklist

      hard-to-understand areas.


app/database/TransactionDao.ts
```diff
@@ -1,11 +1,12 @@
 export const {
     getPending: getPendingTransactions,
     hasPending: hasPendingTransactions,
+    getFilteredPendingTransactions,
     hasEncryptedTransactions,
     hasPendingShieldedBalanceTransfer,
     update: updateTransaction,
     insert: insertTransactions,
-    upsertTransactionsAndUpdateMaxId,
     getTransactionsForAccount: getTransactionsOfAccount,
```
shjortConcordium, line 9: This is now unused, I think. It should be removed.

app/database/migrations/20211108144203_create_decrypted_amounts_table.ts
```diff
@@ -0,0 +1,36 @@
...
+            table.string('amount').notNullable();
+        });
+
+        // Remove the maxTransactionId as we no longer need it when we retrieve
+        // transactions from the wallet proxy, instead of having a local copy.
+        await t.schema.alterTable(accountsTable, (table) => {
+            table.dropColumn('maxTransactionId');
+        });
+
+        // Delete all transactions are we now retrieve transactions from the wallet proxy,
```
shjortConcordium, line 23:
```suggestion
        // Delete all transactions as we now retrieve transactions from the wallet proxy,
```

app/database/migrations/20211108144203_create_decrypted_amounts_table.ts
```diff
@@ -0,0 +1,36 @@
+import { Knex } from 'knex';
+import {
+    decryptedAmountsTable,
+    accountsTable,
+    transactionTable,
+} from '~/constants/databaseNames.json';
+
+export async function up(knex: Knex): Promise<void> {
```
shjortConcordium, line 8: Should this copy the already decrypted amounts from the transactions table into the decryptedAmountsTable?

app/utils/httpRequests.ts
```diff
@@ -19,19 +24,47 @@ export async function httpGet<T>(
...
+        transactionFilter,
+        limit,
+        TransactionOrder.Ascending,
+        id
+    );
+
+    return {
+        transactions: response.transactions,
+        full: response.full,
+    };
```
shjortConcordium, line 48: Is there any reason we are de/reconstructing the output in this and getTransactionsDescending?
```suggestion
    return window.http.getTransactions(
        address,
        transactionFilter,
        limit,
        TransactionOrder.Ascending,
        id
    );
```

app/utils/decryptHelpers.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Decrypts the encrypted transfers in the provided transaction list. This is
+ * done by using the provided PRF key, which has to belong to the corresponding
+ * receiver account.
+ *
+ * Note: If the PRF key and account mismatches, then this method
+ * will run indefinitely.
+ * @param encryptedTransfers the encrypted transfers to decrypt
+ * @param the account that the transactions are for
```
shjortConcordium, line 12:
```suggestion
 * @param accountAddress the account that the transactions are for
```

app/preload/http.ts
```diff
@@ -44,13 +48,39 @@ async function httpsGet<T>(
...
+    );
+    return filteredTransactions;
+}
+
+/**
+ * Loads transactions from the wallet proxy for the account with the provided address. The supplied
+ * transaction filter is applied for reward filtering, and filtering based on dates. All other parts
+ * of the filter is ignored, as the wallet proxy does not support it!
+ * @param address the account address to get transactions for
+ * @param transactionFilter is used to filter the request, however only from/to date, and filters for reward types are currently used
```
shjortConcordium, line 72: This line is somewhat incorrect, because we filter on the remaining types afterwards.

app/preload/accountReport.ts
```diff
@@ -8,16 +8,75 @@ import {
...
+            withDecryptedAmounts.push(t);
+        } else {
+            const amount = decryptedAmounts.find(
+                (decrypted) => decrypted.transactionHash === t.transactionHash
+            )?.amount;
+            if (amount) {
+                withDecryptedAmounts.push({ ...t, decryptedAmount: amount });
+            } else {
+                const decryptedTransaction = (
+                    await decryptTransactions(
```
shjortConcordium, line 61: In this function, it says that there is a "significant performance hit if we decrypt one at a time", and yet we do that here. Was the performance hit not that significant, or should we decrypt these together?

app/pages/Accounts/useAccountSync.ts
```diff
@@ -73,45 +62,61 @@ export default function useAccountSync(onError: (message: string) => void) {
...
+    // - the status of the account changes, e.g. if it is confirmed,
+    // - the user switches between viewing shielded or unshielded transactions
+    // given that the account info has already been loaded.
+    const loadDependencyArray = [
+        JSON.stringify(account?.transactionFilter),
+        account?.status,
+        accountInfoLoaded,
+        viewingShielded,
+    ];
+    useEffect(() => {
```
shjortConcordium, line 113: I was able to make endless load, and duplicate transactions in the list, on a finalizer. Though not consistently, I think it needs to be loading while an accountInfo update triggers a "loadNewTransactions"

app/pages/Accounts/useAccountSync.ts
```diff
@@ -21,38 +20,28 @@ const accountInfoUpdateInterval = 30000;
...
 export const accountInfoFailedMessage = (message: string) =>
     `Failed to load account information from your connected node due to: ${message}`;
-const updateTransactionsFailedMessage = (message: string) =>
-    `Failed to load transactions from external service due to: ${message}`;
 
 /**
- * Keeps account info and transactions for selected account in sync. Is dependant on a full re-mount when chosen account changes.
+ * Ensures that the account info (from the node) and the list of transactions (from the wallet proxy),
+ * for the selected account, are kept synchronized. Is dependant on a full re-mount when the
+ * chosen account changes.
```
shjortConcordium, line 27: I don't know if this is a reasonable condition (or whether it was ever satisfied)

app/pages/Accounts/AccountDetailsPage/AccountReport/AccountReport.tsx
```diff
@@ -56,30 +69,66 @@ export default function AccountReport({ location }: Props) {
...
+                            ...prfKeyMap,
+                        };
+                        updatedPrfKeyMap[address] = {
+                            credentialNumber,
+                            prfKeySeed,
+                        };
+                        prfKeyMap = updatedPrfKeyMap;
+                    }
                 }
+                accountsToReport.push(account);
```
shjortConcordium, line 131: If i don't decrypt, then i get an error that credentialNumber is undefined. (but only the first time)

app/pages/Accounts/AccountDetailsPage/AccountReport/AccountReport.tsx
```diff
@@ -56,30 +69,66 @@ export default function AccountReport({ location }: Props) {
...
+    /**
+     * Constructs one or more account reports. If there are multiple chosen accounts,
+     * then the reports will be bundled into a zip file.
      */
     const makeReport = useCallback(
         async (filters: TransactionFilter) => {
+            let prfKeyMap: Record<string, CredentialNumberPrfKey> = {};
             const accountsToReport: Account[] = [];
             for (const account of accounts) {
-                const hasEncrypted = await containsEncrypted(account, filters);
```
shjortConcordium, line 79: containsEncrypted is no longer used, It seems.

app/features/TransactionSlice.ts
```diff
@@ -82,8 +77,111 @@ let latestLoadingRequestId: string | undefined;
...
+    return transactionsInDescendingOrder;
+}
+
+/**
+ * Enriches any encrypted transfers (with or without memo) with decrypted amounts, if they
+ * have previously been decrypted and had their amounts stored locally in the database.
+ * @param transactions an array of transactions from the wallet proxy
+ * @returns the transactions enriched with decrypted amounts, and a boolean indicating whether we had decrypted amounts for all the encrypted transfers
+ */
+async function enrichWithDecryptedAmounts(
```
shjortConcordium, line 140: It is a little confusing that this and a function in `preload/AccountReport.ts`

app/pages/Accounts/useAccountSync.ts
shjortConcordium: Seems it is (because it is only used directly by withAccountSync)

app/pages/Accounts/useAccountSync.ts
shjortConcordium: Seems I can't do this anymore.

app/pages/Accounts/AccountDetailsPage/AccountReport/AccountReport.tsx
shjortConcordium: I don't get this anymore.

app/database/migrations/20211108144203_create_decrypted_amounts_table.ts
orhoj: That's a good question. The migration can become quite slow if we do not just drop everything. I can try to profile it a little bit before we decide.

app/features/TransactionSlice.ts
orhoj: A little, but as they kind of do the same we will keep it as is.

app/database/migrations/20211108144203_create_decrypted_amounts_table.ts
orhoj: We have decided that it will be okay for people to decrypt again after the upgrade. There will be few users that have encrypted transfers at the point of the release, so we will skip it here.

app/pages/Accounts/useAccountSync.ts
```diff
@@ -124,10 +129,11 @@ export default function useAccountSync(onError: (message: string) => void) {
                 force: true,
             })
         );
+        load.then(() => setIsLoadDone(true)).catch(noOp);
```
shjortConcordium, line 132: Maybe this should actually be a finally? (though that might be a moot point, but I'm not sure when this could throw)

shjortConcordium (approved):

app/pages/Accounts/useAccountSync.ts
orhoj: Instead I have caught errors and use that to display the error.

shjortConcordium (changes requested):
Please don't show an error when loads are aborted.

shjortConcordium:
LGTM

shjortConcordium (approved):
LGTM

PR #163, opened 2021-11-17
Stop updating node status state on no change

orhoj:
## Purpose
I noticed our console was spammed with `setNodeStatus` state updates even when the state was not changing. This makes it so that the state update is only dispatched if it has changed.

## Changes
- Check if the status has changed before updating the status state.
- Start the controller, so that `abort()` actually aborts.

## Checklist
      hard-to-understand areas.


shjortConcordium (approved):
LGTM!

PR #169, opened 2021-11-18
Improve performance of account report

orhoj:
## Purpose
Improve the performance of the account report (significantly).

## Changes
- Avoid using `fromDate` and `toDate` in as many queries to the wallet proxy as possible. Instead, use `from` and paginate over that and do the filtering in-memory. The wallet proxy does not perform well if we use queries with the dates.
- Removed filtering based on `smallestBlockTime` as it resulted in skipping transactions that should not have been skipped.

## Checklist

      hard-to-understand areas.


shjortConcordium (approved):
LGTM!

PR #173, opened 2021-11-22
Fix issues with finding min and max id

orhoj:
## Purpose
The reducer functions used to find `minId` and `maxId` used string comparisons, which lead to bugs where the transaction loading would keep loading the same transactions, instead of getting new transactions. This PR fixes that by calculating the correct `minId` and `maxId`.

## Changes
As the transaction list state is sorted descendingly on the `id` field, the following changes are made to fix the issue:
- `maxId` is be found as the first item in the transaction state with an `id`.
- `minId` is found as the last item in the transaction state with an `id`.

## Checklist
      hard-to-understand areas.

### DIFF
--- a/app/features/TransactionSlice.ts
+++ b/app/features/TransactionSlice.ts
@@ -93,7 +93,7 @@ async function getNewTransactions(
 ): Promise<TransferTransaction[]> {
     // As the transactions in the state are in descending order on their id, the maxId
     // can be found as the first item with an id (if any exist).
-    const maxId = transactionsInState.map((t) => t.id).filter(isDefined)[0];
+    const maxId = transactionsInState.find((t) => t.id !== undefined)?.id;
 
     const transactions: IncomingTransaction[] = [];
     let full = true;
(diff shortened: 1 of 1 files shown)


app/features/TransactionSlice.ts
```diff
@@ -83,21 +83,17 @@ const forceLock = new Mutex();
...
-    const maxId = transactionsInState
-        .map((t) => t.id)
-        .filter(isDefined)
-        .reduce<string | undefined>(
-            (max, cur) => (!max || max < cur ? cur : max),
-            undefined
-        );
+    // As the transactions in the state are in descending order on their id, the maxId
+    // can be found as the first item with an id (if any exist).
+    const maxId = transactionsInState.map((t) => t.id).filter(isDefined)[0];
```
shjortConcordium, line 96: I would suggest using .find() instead of mapping.

app/features/TransactionSlice.ts
```diff
@@ -216,13 +212,12 @@ export const loadTransactions = createAsyncThunk(
             }
         };
 
-        const minId = state.transactions.transactions
+        // The array of transactions is sorted in descending order on the id. Therefore the
+        // minimum id can be found as the final transaction in the array.
+        const transactionIdArray = state.transactions.transactions
```
shjortConcordium, line 217: Here, couldn't we assume that if the last element didn't have an id, then none of them did?

orhoj: We could but it would be more fragile to future changes.

shjortConcordium (approved):

PR #177, opened 2021-11-22
Get identity provider support mail from proxy

shjortConcordium:
## Purpose

Load identity provider support mail from wallet proxy, instead of hardcoding it.

## Changes

Reworked FailedIdentityDetails to load support mail.
Removed fallback, where user were given the provider's url if support mail wasn't available.

## Checklist

      hard-to-understand areas.


app/components/IdentityCard/FailedIdentityDetails.tsx
```diff
@@ -83,58 +89,50 @@ ${getOS()}`;
...
+    if (mail === undefined) {
+        return (
+            <div className={clsx('body3 p20', styles.failedDetails)}>
+                <SimpleErrorModal
+                    show={failedConnect}
+                    header="Unable to load identity provider email. Please check your internet connection."
+                    onClick={() => setFailedConnect(false)}
+                />
+                {failedDetails}
+                <Loading className="pV20" inline />
```
orhoj, line 101: If the loading fails, i.e. due to a missing internet connection, then the "Loading" is shown forever, even though we never do any retry to get the mail.

app/components/IdentityCard/FailedIdentityDetails.tsx
```diff
@@ -83,58 +89,50 @@ ${getOS()}`;
         );
     }
 
+    if (mail === undefined) {
+        return (
+            <div className={clsx('body3 p20', styles.failedDetails)}>
+                <SimpleErrorModal
+                    show={failedConnect}
+                    header="Unable to load identity provider email. Please check your internet connection."
```
orhoj, line 97:
```suggestion
                    header="Unable to load identity provider contact information. Please check your internet connection."
```

app/components/IdentityCard/FailedIdentityDetails.tsx
```diff
@@ -65,6 +68,38 @@ ${getOS()}`;
...
+    const loadSupportMail = useCallback(async () => {
+        let supportMail;
+        try {
+            supportMail = await getIdentityProviderMail(
+                identityProvider.ipInfo.ipIdentity
+            );
+        } catch {
+            // continue regardless of error
+        }
+        setConnecting(false);
```
orhoj, line 88: As we try to update state asynchronously, this has to be cleaned up if the component is unmounted to avoid a leak.

orhoj, line 88: As we try to update state asynchronously, this has to be cleaned up if the component is unmounted to avoid a leak.

orhoj (changes requested):
Fix unmount leak.

app/components/IdentityCard/FailedIdentityDetails.tsx
shjortConcordium: I have added a controller.

orhoj (approved):

PR #178, opened 2021-11-23
Prevent user from sending to an account's own aliases

### LINKED ISSUE #171: Disallow sending transfers to an account's own aliases
**Description**
With the introduction of aliases it is now possible to send transfers to yourself. We have blocked sending to an account's own address, and this change should block sending to an alias of the sending account to align the behavior.

**Steps to Reproduce**
1. Add an address book entry with an alias of an account you have.
2. Send a transfer and notice that the alias shows up and is selectable. Choosing this address book entry will send the transfer to the account itself.

**Expected Result**
An account should not have the option of sending transactions to itself, i.e. the address book entries that contain aliases should not be shown to the user.

orhoj:
## Purpose
A user should not be able to make a transfer from an account to one of its own aliases.

Closes #171 

## Changes
- When selecting a recipient we now filter away any aliases of the sending account.
- When adding an address book entry in the context of creating a transfer (multisignature and single), the user is blocked from adding an alias of the account that is creating that transfer.

## Checklist
      hard-to-understand areas.


shjortConcordium (approved):
LGTM, though I am wondering if we should also block exporting credentials to aliases. (Perhaps just if you have already exported to that account already)

PR #179, opened 2021-11-23
use correct KeyGen for new identities + identity versioning

shjortConcordium:
## Purpose

Use correct version of bls keygen, that has been added to concordium-base.
Add versioning to identities to support old identities, which used the incorrect bls keygen.

## Changes

Also split Identity into sub-types based on status.
Block creating an identity, on an already used index.

## Checklist

      hard-to-understand areas.


app/database/migrations/20211109124703_add_identity_versions.ts
```diff
@@ -0,0 +1,16 @@
+import { Knex } from 'knex';
+import { identitiesTable } from '~/constants/databaseNames.json';
+
+const versionField = 'version';
+
+export async function up(knex: Knex): Promise<void> {
+    return knex.schema.alterTable(identitiesTable, (table) => {
+        table.integer(versionField).defaultTo(0);
```
orhoj, line 8: Please add a 'not null' constraint to the field.

app/features/IdentitySlice.ts
```diff
@@ -40,15 +40,17 @@ export const identitiesSelector = (state: RootState) =>
...
-        (identity: Identity) => IdentityStatus.Confirmed === identity.status
-    );
+    state.identities.identities.filter(isConfirmedIdentity);
 
 export const chosenIdentitySelector = (state: RootState) =>
     state.identities.identities.find(
         (i) => i.id === state.identities.chosenIdentityId
     );
 
+export const specificIdentitySelector = (identityId?: number) => (
```
orhoj, line 50: It seems odd to have a selector for a specific identity with an optional input. I would expect `identityId` to be required for this selector to make sense.

app/pages/Accounts/Accounts.module.scss
```diff
@@ -55,3 +55,16 @@
     align-items: center;
     position: relative;
 }
+
+.verifyAddress {
```
orhoj, line 59: These changes are probably not needed for this update.

app/pages/Identities/IdentityView.tsx
```diff
@@ -17,12 +17,9 @@ export default function IdentityView() {
 
     return (
         <>
-            <FailedIdentityModal
-                identityId={identity.id}
-                identityName={identity.name}
-                isRejected={identity.status === IdentityStatus.Rejected}
-                detail={identity.detail}
-            />
+            {isRejectedIdentity(identity) && (
```
orhoj, line 20: The `FailedIdentityModal` already does a check on whether `identity` is rejected. Do we need both checks?

app/pages/IdentityIssuance/PickProvider/PickProvider.tsx
```diff
@@ -180,6 +220,7 @@ export default function IdentityIssuanceChooseProvider({
                     ledgerCallback={withLedger}
                     preCallback={getIdentityNumber}
                     disabled={!provider}
+                    identityVersion={0}
```
orhoj, line 223: Why does this have to be hardcoded to `0`? I would expect that we would use `1` for new identities.

app/pages/Recovery/PerformRecovery.tsx
```diff
@@ -33,7 +33,8 @@ async function getPrfKeySeed(
     identityNumber: number
 ) {
     setMessage('Please allow recovering credentials');
-    const prfKeySeed = await ledger.getPrfKeyRecovery(identityNumber);
+    // Version = 0, because Recovery always uses the seed, to check both versions.
```
orhoj, line 36: I do not understand this comment. If `getPrfKeyRecovery` always returns the seed, then why would it take the `version` as input?

app/pages/Recovery/util.ts
```diff
@@ -39,11 +39,13 @@ export function getRecoveredIdentityName(identityNumber: number) {
  * Creates an placeholder identity,
  * @param walletId the wallet on which the identity is created.
  * @param identityNumber the identity's number on the wallet.
+ * @param identityVersion the version, which the identity's PRF key and IdCredSec was generated with.
```
orhoj, line 42:
```suggestion
 * @param identityVersion the version, which the identity's PRF key and IdCredSec were generated with.
```

app/utils/identityHelpers.ts
```diff
@@ -4,8 +4,15 @@ import {
...
     AttributeKey,
     AttributeKeyName,
+    PendingIdentity,
+    IdentityStatus,
+    ConfirmedIdentity,
+    RecoveredIdentity,
+    RejectedIdentity,
 } from './types';
 
+export const currentIdentityVersion = 1;
```
orhoj, line 14: We should document somewhere what the two versions mean. Something like:
```
// The identity was created using the legacy BLS12-381 key generation, decoding the seed
// as UTF-8.
identityVersion = 0

// The identity was created using the corrected BLS12-381 key generation, decoding the seed
// as hex.
identityVersion = 1
```
I am not sure if it should be here though.

app/utils/types.ts
```diff
@@ -106,19 +106,42 @@ export enum IdentityStatus {
 /**
  * This Interface models the structure of the identities stored in the database.
```
orhoj, line 107: This comment no longer matches the interface.

app/utils/rustInterface.ts
```diff
@@ -24,45 +24,51 @@ import { getDefaultExpiry, secondsSinceUnixEpoch } from './timeHelpers';
 import { getAccountPath } from '~/features/ledger/Path';
 import { stringify, parse } from './JSONHelper';
 import CredentialInfoLedgerDetails from '~/components/ledger/CredentialInfoLedgerDetails';
+import { currentIdentityVersion } from './identityHelpers';
 
 const rawWorker = new RustWorker();
 const worker = new PromiseWorker(rawWorker);
 
+const versionUsesDeprecatedKeyGen = (v: number) => v === 0;
```
orhoj, line 32: Perhaps `identityCreatedUsingDeprecatedKeyGen` would be more saying.

app/utils/rustInterface.ts
```diff
@@ -24,45 +24,51 @@ import { getDefaultExpiry, secondsSinceUnixEpoch } from './timeHelpers';
...
+
 /**
- * Returns the PrfKey and IdCredSec seeds for the given identity.
+ * Returns the PrfKey and IdCredSec for the given identity.
+ * identityVersion 0 gives the seeds, and 1 gives the BLS keys.
```
orhoj, line 36: I don't think that this is true.

orhoj (changes requested):
Leave a comment

app/pages/Identities/IdentityView.tsx
shjortConcordium: Not really, but I wanted no distinction between rejected and rejectedButWarned identities outside of `FailedIdentityModal`.
So this check is for the type checker, and so that the `FailedIdentityModal` logic only has to run for a failed identity.
The check inside is whether it is rejected or rejectedButWarned, which decides whether the modal should pop up.

app/pages/IdentityIssuance/PickProvider/PickProvider.tsx
shjortConcordium: This is the version to give to the ledger, and for identity issue we have to get the seed always, because we check whether the index has already been used by any of the versions.

app/pages/Recovery/PerformRecovery.tsx
shjortConcordium: I guess it shouldn't take the version.

app/utils/identityHelpers.ts
shjortConcordium: I have created a type `IdentityVersion`, and documented it there.

app/utils/types.ts
shjortConcordium: I moved it down to Identity again.

shjortConcordium:
> Leave a comment

I removed the versioning from the ledger interaction, because it was just clutter here.

app/features/IdentitySlice.ts
shjortConcordium: Yea, it was a convenience thing, because I used it in the account report, where the account is not defined from the start. (But I have cleaned up the versioning, so it was not necessary anymore).

orhoj:
It appears that changing the identity version of an identity does not change the PRF-key used.

orhoj (approved):
LGTMFO

PR #181, opened 2021-11-25
Fix transaction loading

orhoj:
## Purpose
In some cases not all transactions were successfully loaded from the wallet proxy. This change updates the loading of transactions from the wallet proxy, so that we keep paginating until a full page is loaded, or until all transactions are loaded. This should ensure that we will not have missing transactions.

## Changes
- Use transaction filter when loading shielded transactions. This also means that the wallet proxy will not be queried for reward transactions when we load shielded transactions.
- Filter away incoming shielded transactions when showing normal balance.
- When loading transactions keep loading until a full page (after filtering) has been received.
- Only load new transactions (on balance change, encrypted balance change) if no `toDate` has been set.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/features/TransactionSlice.ts
+++ b/app/features/TransactionSlice.ts
@@ -27,6 +27,8 @@ import {
     TransferTransactionWithNames,
     Account,
     IncomingTransaction,
+    TransactionFilter,
+    BooleanFilters,
 } from '../utils/types';
 import {
     isShieldedBalanceTransaction,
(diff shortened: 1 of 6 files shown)


app/features/TransactionSlice.ts
```diff
@@ -85,38 +137,40 @@ const forceLock = new Mutex();
  * @param account the account to get new transactions for
  * @param transactions the current transactions in the state, assumed to be sorted descendingly
  * @param limit the maximum number of transactions to ask the wallet proxy for
+ * @param loadShielded whether the transactions that should be retrieved are shielded transactions or not
```
shjortConcordium, line 140: This name / comment is a bit misleading, I think it should be more clear, that it _only_ shielded transactions.

app/pages/Accounts/useAccountSync.ts
```diff
@@ -70,35 +65,43 @@ export default function useAccountSync(onError: (message: string) => void) {
...
+        if (
+            loadIsDone &&
+            !viewingShielded &&
+            !account?.transactionFilter.toDate
+        ) {
             const load = dispatch(
-                loadNewTransactions({
-                    showLoading: true,
-                })
+                loadNewTransactions({ loadNewShielded: false })
```
shjortConcordium, line 95: These effects could use a shared function

app/features/TransactionSlice.ts
orhoj: I have change all of them to something prefixed with `only`.

shjortConcordium (approved):
LGTM

PR #182, opened 2021-11-26
Fix balance button hitbox

shjortConcordium:
## Purpose

Fix that in the account view, You could only click on the balance button's _balan_ part.

## Changes

Gave the viewing shielded buttons a higher z-index.


orhoj (approved):
Update of the year.

PR #183, opened 2021-11-26
Bugfix: Shielded balance becomes locked after making a shielded transfer

shjortConcordium:
## Purpose

Fix bug, which caused shielded balance to become locked after sending shielded transfer.

## Changes

Saves decrypted amount on finalization.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/utils/TransactionStatusPoller.ts
+++ b/app/utils/TransactionStatusPoller.ts
@@ -11,7 +11,6 @@ import {
     UpdateAccountCredentials,
     Transaction,
     TransferTransaction,
-    TransactionKindString,
 } from './types';
 import {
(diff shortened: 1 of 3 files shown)


app/utils/TransactionStatusPoller.ts
```diff
@@ -68,6 +69,12 @@ function ShieldedTransferConsequence(
     const { newSelfEncryptedAmount, remainingDecryptedAmount } = parse(
         transaction.encrypted
     );
+    if (transaction.decryptedAmount) {
```
orhoj, line 72: This should not insert for shielding and unshielding.

orhoj (approved):

PR #192, opened 2021-12-03
Add Register data transaction

shjortConcordium:
## Purpose

Add support for the register data transaction.
(Companion to https://github.com/Concordium/concordium-ledger-app/pull/36)

## Changes

Add flows for creating register data transactions.
Introduced a common `chooseExpiry` component for multisig baker flows and register data. (The merge with the new baker flows removed the use of `chooseExpiry` from the baker flows (except restake)) 
Fixed bug where cost of memo was slightly underestimated.
Changed `multisig/AccountTransactions/CreateTransaction` to use `CreateTransactionInput` instead of specific fields.

## Checklist

      hard-to-understand areas.


orhoj:
The _Data registration_ item in the transaction log does not display the data that was posted. Is that intentional? Or should it be viewable as a memo for a transfer?

orhoj:
There is a bug (either in the desktop wallet, or in the Ledger code) where it does not display the final character on the Ledger, but instead shows the initial data again. It can be provoked with the following input:
```
01=4601=4395=1035l.13-=vo,1=035o13[p5ovp13[5c[135kco;sj;fkjas;dligjifdagafdgfapdgkl;dfgk;dflgdf;kgjadk;lfhjfdlkhjieofau94yu224i0k,3[w9r-viymwrymbuwriybwirepymvirpeyumviwper,yvuiwperu bwnipermvyrwepujgifdajg;lfdijgdsakajsglkasdjglk;dfhg;dafhglfdjaghalueiw
```
It probably just have to be max length for it to happen, but I did it with the above.

app/features/ledger/Transfer.ts
```diff
@@ -433,6 +437,55 @@ async function signUpdateBakerRestakeEarnings(
...
+            0xe0,
+            INS_REGISTER_DATA,
+            p1,
+            p2,
+            Buffer.from(chunk)
+        );
+    }
+
+    if (!response) {
+        throw new Error('Unexpected missing response from ledger;');
```
orhoj, line 484:
```suggestion
        throw new Error('Unexpected missing response from Ledger');
```

app/pages/multisig/AccountTransactions/MultisignatureAccountTransactions.module.scss
```diff
@@ -1,44 +1,2 @@
 @import 'config/bundle';
 @import '../common/MultiSignatureFlowPage.module.scss';
```
orhoj, line 2: Does it make sense to keep this file now that it only contains other imports?

app/utils/transactionHelpers.ts
```diff
@@ -157,6 +158,30 @@ export function createSimpleTransferWithMemoTransaction(
     });
 }
 
+/**
+ *  Constructs a, register data, transaction object,
```
orhoj, line 162:
```suggestion
 *  Constructs a register data transaction object.
```

orhoj:
Should register data have a single signer flow?

shjortConcordium:
> There is a bug (either in the desktop wallet, or in the Ledger code) where it does not display the final character on the Ledger, but instead shows the initial data again. It can be provoked with the following input:
> 
> ```
> 01=4601=4395=1035l.13-=vo,1=035o13[p5ovp13[5c[135kco;sj;fkjas;dligjifdagafdgfapdgkl;dfgk;dflgdf;kgjadk;lfhjfdlkhjieofau94yu224i0k,3[w9r-viymwrymbuwriybwirepymvirpeyumviwper,yvuiwperu bwnipermvyrwepujgifdajg;lfdijgdsakajsglkasdjglk;dfhg;dafhglfdjaghalueiw
> ```
> 
> It probably just have to be max length for it to happen, but I did it with the above.

Yeah, this is a bug in the ledger app. :see_no_evil: 
I don't check if the first batch is everything or not, so it always displays that. (it should not display if the data is full length)
And the second time it displays everything (including the final characters).

shjortConcordium:
> The _Data registration_ item in the transaction log does not display the data that was posted. Is that intentional? Or should it be viewable as a memo for a transfer?

I did not display it because, unlike with the memos, the wallet proxy does not provide the data.

app/pages/multisig/AccountTransactions/MultisignatureAccountTransactions.module.scss
shjortConcordium: Not really, I changed the users to use `../common/MultiSignatureFlowPage.module.scss` instead.

orhoj:
> > The _Data registration_ item in the transaction log does not display the data that was posted. Is that intentional? Or should it be viewable as a memo for a transfer?
> 
> I did not display it because, unlike with the memos, the wallet proxy does not provide the data.

I think this is an issue as the user cannot see what they actually registered on the chain from within the wallet. I guess it is not available in the wallet-proxy as the mobile wallets do not support the transaction, but we kind of need it. Can you talk to Aleš about whether we could have it added?

The alternative would be that we would have to look it up in the node, but that would probably be quite slow.

shjortConcordium:
> > > The _Data registration_ item in the transaction log does not display the data that was posted. Is that intentional? Or should it be viewable as a memo for a transfer?
> > 
> > 
> > I did not display it because, unlike with the memos, the wallet proxy does not provide the data.
> 
> I think this is an issue as the user cannot see what they actually registered on the chain from within the transaction log. I guess it is not available in the wallet-proxy as the mobile wallets do not support the transaction, but we kind of need it. Can you talk to Aleš about whether we could have it added?
> 
> The alternative would be that we would have to look it up in the node, but that would probably be quite slow.

I talked to Aleš, he says they will add it to the wallet-proxy.

shjortConcordium:
So I've added the display of the data, with the new wallet-proxy.
I'm just reusing the memo field in the database, I don't know if we should do a migration and change the name of that to dataBlob? 
:thinking:

app/components/Transfers/DisplayRegisterData.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { RegisterData } from '~/utils/types';
+import styles from './transferDetails.module.scss';
```
soerenbf, line 3: Please ensure styles are imported last

app/pages/multisig/AccountTransactions/proposal-details/RegisterDataProposalDetails.tsx
```diff
@@ -0,0 +1,28 @@
...
+    estimatedFee,
+    expiryTime,
+    data,
+}: Props) {
+    return (
+        <Details>
+            <AccountDetail title="Account" value={account} first />
+            <DisplayEstimatedFee estimatedFee={estimatedFee} />
+            <PlainDetail title="Data" value={data} />
+            <DisplayTransactionExpiryTime expiryTime={expiryTime} />
```
soerenbf, line 25: Would be nice if this showed "To be determined" when not specified yet. Either that, or the default value (I prefer the first option though).

soerenbf (changes requested):
Some minor stuff, otherwise LGTM

app/pages/multisig/AccountTransactions/proposal-details/RegisterDataProposalDetails.tsx
shjortConcordium: I just I prefer the first option as well.

app/components/Transfers/DisplayRegisterData.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { RegisterData } from '~/utils/types';
+import { useAccountName } from '~/utils/dataHooks';
+import DisplayTransactionExpiryTime from '../DisplayTransactionExpiryTime/DisplayTransactionExpiryTime';
```
soerenbf, line 4:
```suggestion
import DisplayTransactionExpiryTime from '../DisplayTransactionExpiryTime';
```

soerenbf (approved):

PR #198, opened 2021-12-15
Update default node

orhoj:
## Purpose
Use Virtual Hive's public node as the default node in the desktop wallet. This should ease the use of the desktop wallet for users who do not wish to run their own node.

## Changes
- Migrated any default node settings from `127.0.0.1` to `concordiumwalletnode.com`. The idea is that users who have already changed their settings probably do not want it forcibly updated.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -3,9 +3,9 @@ name: Build, lint and test
 on:
     # Triggers the workflow on push or pull request events but only for the main branch
     push:
-        branches: [main]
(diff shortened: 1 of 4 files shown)


shjortConcordium:
Can you do so that testnet builds don't default to this mainnet node?

orhoj:
> Can you do so that testnet builds don't default to this mainnet node?

The migration now only does something if part of a mainnet build.

shjortConcordium (approved):

PR #215, opened 2022-01-14
Fix unresponsive input fields after identityIssuance is aborted.

### LINKED ISSUE #201: Text fields are unresponsive after ID flow is aborted
**Bug Description**
After an ID flow is aborted, you cannot enter text in any text field (e.g. new account/identity name, node address in node settings). DW needs restart. Note that this was pbserved after aborting a hanging flow with DTS. It might not occur when aborting a flow with NB.

**Steps to Reproduce**
- create identity with DTS
--> flow will hang eventually
- abort flow by clicking X
- try to enter text in any text field

**Expected Result**
Text can be entered, e.g. issuance of a new identity can be initiated. 

**Actual Result**
All text fields are unresponsive.

**Versions**
* desktop wallet 1.3.0 beta 2 for stagenet
* iPhone 11, iOS 15.1

shjortConcordium:
## Purpose

Fix #201

## Changes

Instead of using the Prompt component to prompt user, use electron's messageBox,
because the Prompt component blocks the renderer thread, which seems to cause the issue with the input fields.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/pages/IdentityIssuance/NavigationBlock.tsx
+++ b/app/pages/IdentityIssuance/NavigationBlock.tsx
@@ -1,0 +1,49 @@
+import React, { useState, useCallback } from 'react';
+import { useDispatch } from 'react-redux';
+import { push } from 'connected-react-router';
+import { Location } from 'history';
+import { Prompt } from 'react-router-dom';
+
+interface Props {
+    shouldPrompt: (location: Location) => boolean;
+}
+
+const message = 'You are about to abort creating an identity. Are you sure?';
+
+/**
+ * This Component is used to block navigation away from the identity issuance flow.
+ * Uses Prompt to block navigation, and eletron's own messageBox to prompt the user.
+ */
+export default function BlockingPage({ shouldPrompt }: Props) {
+    const [blocking, setBlocking] = useState(true);
+    const dispatch = useDispatch();
+
+    const showPrompt = useCallback(
+        async (location: Location) => {
+            setBlocking(false);
+            const { response } = await window.messageBox({
+                message,
+                type: 'warning',
+                noLink: true,
+                buttons: ['abort', 'cancel'],
+            });
+            if (response === 0) {
+                // 0 = abort, because the response is the index of the pressed button.
+                dispatch(push(location));
+            } else {
+                setBlocking(true);
+            }
+        },
+        [message]
+    );
+
+    const handleNavigation = (nLocation: Location) => {
+        if (!shouldPrompt(nLocation)) {
+            return true;
+        }
+        showPrompt(nLocation);
+        return false;
+    };
+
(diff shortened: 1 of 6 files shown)


app/pages/IdentityIssuance/NavigationBlock.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+interface Props {
+    shouldPrompt: (location: Location) => boolean;
+}
+
+const message = 'You are about to abort creating an identity. Are you sure?';
+
+/**
+ * This Component is used to block navigation away from the identity issuance flow.
+ * Uses Prompt to block navigation, and eletron's own messageBox to prompt the user.
```
orhoj, line 15:
```suggestion
 * Uses Prompt to block navigation, and Electron's own messageBox to prompt the user.
```

app/pages/IdentityIssuance/NavigationBlock.tsx
```diff
@@ -0,0 +1,53 @@
...
+    const dispatch = useDispatch();
+
+    const showPrompt = useCallback(
+        async (location: Location) => {
+            setBlocking(false);
+            const { response } = await window.messageBox({
+                message,
+                type: 'warning',
+                noLink: true,
+                buttons: ['abort', 'cancel'],
```
orhoj, line 28:
```suggestion
                buttons: ['Abort', 'Cancel'],
```

app/pages/IdentityIssuance/NavigationBlock.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * Uses Prompt to block navigation, and eletron's own messageBox to prompt the user.
+ */
+export default function NavigationBlock({ shouldPrompt }: Props) {
+    const [blocking, setBlocking] = useState(true);
+    const dispatch = useDispatch();
+
+    const showPrompt = useCallback(
+        async (location: Location) => {
+            setBlocking(false);
+            const { response } = await window.messageBox({
```
orhoj, line 24: The title of the message box is `Electron`. It should be changed.

shjortConcordium: It is only Electron in development, in production it is "Concordium wallet". 

I'm not sure if that is okay or I should explicitly set it to something else?

app/pages/IdentityIssuance/NavigationBlock.tsx
```diff
@@ -0,0 +1,54 @@
...
+                message,
+                type: 'warning',
+                noLink: true,
+                title: 'Concordium',
```
orhoj, line 28: Since we are renaming the wallet, I think it would be better if it just matched the full name?
```suggestion
                title: 'Concordium Desktop Wallet',
```

orhoj (approved):

PR #259, opened 2022-04-08
Baker/delegation multisig flows from accounts fix start page

### LINKED ISSUE #256: Multi signature transaction - wrong routing to account selection
**Bug Description**
In attempt to update the pool setting from open-for-all to close-for-new DW routes to Stake settings page instead of Pool settings

**Steps to Reproduce**
1. Set up account with multi-sig transactions.
2. Register a baker and open a pool.
3. Select account -> Baking

**Expected Result**
Pool settings page should open.

**Actual Result**
User needs to select account again even if the starting point was the actual account.

**Versions**
- Software Version: V 1.4.0-alpha.0

shjortConcordium:
Starting the new Baker/delegation flows from accounts page with a multisig accounts incorrectly starts at account selection.
Fix #256


soerenbf (approved):
Nice!

PR #261, opened 2022-04-08
Improve pickAmount

### LINKED ISSUE #257: Auto focus on the amount doesn't erase decimal point
**Bug Description**
If user wants to send CCDs, selecting Send CCDs opens Transaction proposal page and autofocus is on the amount which is 0.00.
If user just starts to type, amount that user wants to insert will just start after 0.00. This can be somewhat inconvenient for the user because user needs to erase the input manually afterwards to start without the decimal point. On the other hand, behavior is inconsistent because if user clicks in the field before writing (while value is still 0.00) the whole value will be cleared and user can insert values without decimal point.

**Steps to Reproduce**
1. Select account that has some CCDs.
2. Transaction proposal page will open, notice 0.00 is in focus.
3. Attempt to insert 1.

**Expected Result**
Inserted value is 1 CCD. I assume that solution is to clear the input (0.00) right after auto-focus?

**Actual Result**

1. It seems that validation is active right after autofocus, which makes showing validation message 0.00 is not a valid input even if user didn't inserted anything yet.
2. After just inserting 1, inserted value is 0.001 CCD.
3. This is true also for staking/restaking. I assume this is true for every page that contains insert field for amount of CCDs.

**Versions**
- Software Version: V 1.4.0-alpha.0

https://user-images.githubusercontent.com/98821384/161973484-2180adfe-d359-471b-8c86-7f8fa7799d04.mp4

shjortConcordium:
## Purpose

Fix #257

## Changes

- `InlineNumber` can now clear default on focus instead of on click. 
- Improved `PickAmount` error handling so it doesn't instantly give an error.
- Renamed `GtuInput` to `CcdInput`.

## Checklist

      hard-to-understand areas.


app/components/Form/InlineNumber/InlineNumber.tsx
```diff
@@ -177,18 +178,25 @@ export default function InlineNumber({
...
-        setIsFocused(true);
-        onFocus();
-    }, [onFocus]);
+    const handleFocus: FocusEventHandler<HTMLInputElement> = useCallback(
+        (e) => {
+            setIsFocused(true);
+            if (clearOnFocus && e.currentTarget.value === formattedFallback) {
+                setInnerValue('');
+                e.currentTarget.style.width = '6px';
+            }
```
soerenbf, line 195: Maybe we should make extract this into a function, now that it's used twice?

soerenbf: Although... can't we just remove the onclick handler now? The focus handler should also be executed when the input is clicked, as it also triggers focus?

soerenbf (approved):

PR #265, opened 2022-04-08
Transaction expiry time error

### LINKED ISSUE #251: "Transaction expiry time must be before the effective time" message doesn't (dis)appear when it should.
**Bug Description**

When updating parameters in the desktop wallet, we have to give values in the fields "Effective time" and "Transaction expiry time" such that the effective time is after the transaction expiry time. If the transaction expiry time is changed to after the effect time, then an error message saying "Transaction expiry time must be before the effective time" appears. And if this message is on screen, and the time is changed back to something before the effective time, the message disappears. But the same does not happen if it is the effective time that is changed instead of the transaction expiry time. This results in invalid times but no error message, or an error message when the times are valid. (Note that clicking on continue works fine if the values are good and will display the error message if the values are wrong.)

**Steps to Reproduce**

- Select any proposal for updating parameters in the desktop wallet.
- Change the effective time to something before the transaction expiry time. No error message appears. (We expect one to appear.)
- Change the transaction expiry time to something after the effective time. The error message appears. Change the effective time to something after the expiry time. The error message doesn't disappear. (We expect it to disappear.)

**Versions**
- Mac OS 12.3.1
- Concordium Desktop Wallet stagenet Version 1.4.0-alpha.0 (1.4.0-alpha.0)

soerenbf:
## Purpose

Fixes #251 

## Changes

- Ensure onblur event is triggered for datepicker when changing value.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/pages/multisig/MultiSignatureCreateProposal/MultiSignatureCreateProposal.tsx
+++ b/app/pages/multisig/MultiSignatureCreateProposal/MultiSignatureCreateProposal.tsx
@@ -130,7 +130,7 @@ function MultiSignatureCreateProposal({
                     render={() => (
                         <MultiSignatureLayout
                             pageTitle={handler.title}
-                            delegateScroll
+                            delegateScroll={isKeyUpdate}
                         >
                             <BuildComponent
                                 type={type}
--- a/app/components/Form/DatePicker/DatePicker.tsx
+++ b/app/components/Form/DatePicker/DatePicker.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-duplicates */
 import clsx from 'clsx';
-import React from 'react';
+import React, { useCallback } from 'react';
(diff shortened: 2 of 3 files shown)


app/pages/multisig/MultiSignatureCreateProposal/MultiSignatureCreateProposal.tsx
```diff
@@ -130,7 +130,7 @@ function MultiSignatureCreateProposal({
                     render={() => (
                         <MultiSignatureLayout
                             pageTitle={handler.title}
-                            delegateScroll
+                            delegateScroll={isKeyUpdate}
```
shjortConcordium, line 133: What is this "little" change you sneaked in.

shjortConcordium (approved):

PR #289, opened 2022-05-02
Show a warning when baker stake exceeds the capital bound

### LINKED ISSUE #283: Provide warning to the baker if the stake will breach capital bound
**Description**
It is possible to stake amount of CCDs that will breach the capital bound without any feedback. All the stake that is over the bound is not counted the usual way when calculating staking rewards.

**Steps to Reproduce**
Stake the amount that is larger of x% of all staked capital.
x is capitalBound parameter and on the Stagenet it is currently set to 25%.

**Expected Result**
Warning that will inform the user about the capital bound.

**Actual Result**
There is no warning.

**Versions**
- Software Version: V 1.4.0-beta.0

shjortConcordium:
## Purpose

Fix #283

## Changes

Display a warning, and the estimated limit, when breaching the capitalBound.
Created a hook for getting capitalBound.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/components/PickBakerStakeAmount.tsx
+++ b/app/components/PickBakerStakeAmount.tsx
@@ -42,7 +42,7 @@ interface Props {
 function useCapitalBoundCheck(
     accountInfo: AccountInfo | undefined,
     stake: string
-) {
+): { showWarning: true; limitAfterUpdate: bigint } | { showWarning: false } {
     const rewardStatus = useAsyncMemo(getRewardStatusLatest);
     const capitalBound = useCapitalBound();
 
@@ -52,7 +52,7 @@ function useCapitalBoundCheck(
         !rewardStatus ||
         !isRewardStatusV1(rewardStatus)
     ) {
-        return { showWarning: false, limitAfterUpdate: 0n };
+        return { showWarning: false };
     }
     const newStake = ccdToMicroCcd(stake);
     const currentStake =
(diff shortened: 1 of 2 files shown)


app/components/PickBakerStakeAmount.tsx
```diff
@@ -24,6 +36,38 @@ interface Props {
...
+    const rewardStatus = useAsyncMemo(getRewardStatusLatest);
+    const capitalBound = useCapitalBound();
+
+    if (
+        !capitalBound ||
+        !isValidCcdString(stake) ||
+        !rewardStatus ||
+        !isRewardStatusV1(rewardStatus)
+    ) {
+        return { showWarning: false, limitAfterUpdate: 0n };
```
orhoj, line 55: You can type this so that `limitAfterUpdate` only has to have a value if `showWarning === true`. Then you only have to check for `showWarning` when using it. Also it improves readability as `limitAfterUpdate` is not really `0` in this case.

orhoj (changes requested):

app/components/PickBakerStakeAmount.tsx
```diff
@@ -24,6 +36,41 @@ interface Props {
...
+        return { showWarning: false };
+    }
+    const newStake = ccdToMicroCcd(stake);
+    const currentStake =
+        accountInfo && isBakerAccount(accountInfo)
+            ? accountInfo.accountBaker.stakedAmount
+            : 0n;
+    const newTotalStake =
+        rewardStatus.totalStakedCapital - currentStake + newStake;
+    const limitAfterUpdate =
```
orhoj, line 64: This limit calculation has to also consider any delegated amounts.

shjortConcordium: No, the limit has nothing to do it. I have fixed it, so the delegated amount is taken into consideration when checking if the bound has been violated.

orhoj (approved):

PR #293, opened 2022-05-03
285 notify user of closing baker pool

orhoj:
## Purpose
Notify users (at startup) if they are currently delegating to a baker pool that is shutting down.

## Changes
- At startup the application checks whether any delegating account is delegating to a pool that is in the process of closing. A notification is created for each account where this is the case.
- The delegation view now shows if the target pool is shutting down and includes when it will happen.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/components/ClosingBakerPoolNotification.tsx
+++ b/app/components/ClosingBakerPoolNotification.tsx
@@ -25,7 +25,9 @@ export default function ClosingBakerPoolNotification({
             <div className="flex alignCenter">
                 The target pool of the following account will close soon.
             </div>
-            <div className="flex alignCenter mT10">{accountName}</div>
+            <div className="flex alignCenter mT10">
+                <b>{accountName}</b>
+            </div>
             <div className="inlineFlexColumn mT20">
                 <Button className="mT10" size="tiny" onClick={onClose}>
                     Remind me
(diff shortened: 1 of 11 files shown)


app/components/Notification/Notification.module.scss
```diff
@@ -8,6 +8,7 @@
     background-color: $color-white;
     width: 100%;
     height: 160px;
+    z-index: 1000;
```
shjortConcordium, line 11: Doesn't seem to work for me

app/components/ClosingBakerPoolNotification.tsx
```diff
@@ -0,0 +1,36 @@
...
+    return (
+        <Notification
+            className="flexColumn alignCenter"
+            level={NotificationLevel.ClosingBakerPool}
+            onCloseClick={onClose}
+        >
+            <div className="flex alignCenter">
+                The target pool of the following account will close soon.
+            </div>
+            <div className="flex alignCenter mT10">{accountName}</div>
```
shjortConcordium, line 28: Maybe make this bold to make it more obvious that it is an account name :upside_down_face:

shjortConcordium (approved):

PR #321, opened 2022-06-10
Add support for Ledger Nano S Plus

orhoj:
## Purpose
Support the Ledger Nano S Plus.

## Changes
- Updated the dependency used to communicated with the Ledgers (`@ledgerhq/hw-transport-node-hid-singleton`), to one that also supports the Ledger Nano S Plus.
- Disabled the new auto disconnect functionality that was added by Ledger to keep connection behavior as it was prior to the dependency update.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.5.0
+
+### Added
+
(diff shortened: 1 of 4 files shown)


orhoj:
We need to consider handling the type of device as we store all as nanos in the database.

PR #339, opened 2022-08-03
Delegation link

orhoj:
## Purpose
When delegating to a baker we want to provide help so that the user can find a nice pool to join.

## Changes
- Now links to ccdscan/staking where baker pools can be found.
- Fixed the download page link in manual update notifications (rpm / deb) to point to mainnet instead of testnet.

## Checklist
      hard-to-understand areas.

### DIFF
--- a/app/components/ManualUpdateNotification.tsx
+++ b/app/components/ManualUpdateNotification.tsx
@@ -10,7 +10,7 @@ interface Props {
 }
 
 const downloadUrl =
-    'https://developer.concordium.software/en/testnet/net/installation/downloads.html#concordium-desktop-wallet';
+    'https://developer.concordium.software/en/mainnet/net/installation/downloads.html#concordium-desktop-wallet';
 
 export default function ManualUpdateNotification({ version, onClose }: Props) {
     return (
--- a/app/components/ManualUpdateNotification.tsx
+++ b/app/components/ManualUpdateNotification.tsx
@@ -3,14 +3,12 @@ import CogIcon from '@resources/svg/settings.svg';
 import Notification from './Notification';
 import { NotificationLevel } from '~/features/NotificationSlice';
 import Button from '~/cross-app-components/Button';
+import urls from '~/constants/urls.json';
 
 interface Props {
     version: string;
     onClose(): void;
 }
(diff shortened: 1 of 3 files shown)


app/components/ManualUpdateNotification.tsx
```diff
@@ -10,7 +10,7 @@ interface Props {
 }
 
 const downloadUrl =
-    'https://developer.concordium.software/en/testnet/net/installation/downloads.html#concordium-desktop-wallet';
+    'https://developer.concordium.software/en/mainnet/net/installation/downloads.html#concordium-desktop-wallet';
```
shjortConcordium, line 13: Maybe this should just be moved to `constants/urls`?

app/components/Transfers/configureDelegation/DelegationTargetPage.tsx
```diff
@@ -128,24 +128,32 @@ export default function DelegationTargetPage({
...
-                            about finding one here:
+                            If you don&apos;t already know which baker pool you
+                            want to delegate an amount to, you can look for one
+                            here:
+                        </p>
+                        <p className="mV0">
+                            <ExternalLink href={urls.ccdScanStakingPage}>
+                                CCDScan.io/staking
+                            </ExternalLink>
                         </p>
```
shjortConcordium, line 139: I guess it could link to the testnet one for the testnet wallet, but that is probably overkill for this.

shjortConcordium (approved):

app/components/Transfers/configureDelegation/DelegationTargetPage.tsx
orhoj: No I think we should do that, we just didn't do that before.

orhoj: But note that we just use the testnet link for everything not mainnet. We don't have ccdscan for stagenet as far as I know.

shjortConcordium: We have it at:
https://stagenet.ccdscan.io/staking :shrug:

PR #351, opened 2023-09-21
Fix identity issuance and recovery errors.

shjortConcordium:
## Purpose

Unblock identity issuance and recovery.

## Changes

- Change accountInfo checks to handle error thrown when account does not exist.
- Fixed some types.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/components/PickBakerStakeAmount.tsx
+++ b/app/components/PickBakerStakeAmount.tsx
@@ -45,7 +45,7 @@ function useCapitalBoundCheck(
     accountInfo: AccountInfo | undefined,
     stake: string
 ): { showWarning: true; limitAfterUpdate: bigint } | { showWarning: false } {
-    const rewardStatus = useAsyncMemo(getRewardStatusLatest);
+    const rewardStatus = useAsyncMemo(getRewardStatusLatest, noOp, []);
     const poolStatus = useAsyncMemo(
         async () => {
             if (accountInfo) {
--- a/app/pages/IdentityIssuance/PickProvider/PickProvider.tsx
+++ b/app/pages/IdentityIssuance/PickProvider/PickProvider.tsx
@@ -67,7 +67,7 @@ async function hasPrfKeySeedBeenUsed(
     credId = await computeCredIdFromSeed(prfKeySeed, 0, global, 1);
     try {
         info = await getAccountInfoOfCredential(credId, blockHash);
-        return Boolean(info);
+        return true;
     } catch {
         return false;
     }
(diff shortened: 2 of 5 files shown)


app/pages/IdentityIssuance/PickProvider/PickProvider.tsx
```diff
@@ -54,14 +54,23 @@ async function hasPrfKeySeedBeenUsed(
...
+    } catch {
+        // The seed was not used with version 0;
     }
     // Check if this index have been used with version 1 keygen.
     credId = await computeCredIdFromSeed(prfKeySeed, 0, global, 1);
-    info = await getAccountInfoOfCredential(credId, blockHash);
-    return Boolean(info);
+    try {
+        info = await getAccountInfoOfCredential(credId, blockHash);
+        return Boolean(info);
```
orhoj, line 70: It doesn't make sense to do `Boolean(info)` when `info` can't be `undefined`.
```suggestion
        return true;
```

app/components/PickBakerStakeAmount.tsx
```diff
@@ -45,7 +45,7 @@ function useCapitalBoundCheck(
     accountInfo: AccountInfo | undefined,
     stake: string
 ): { showWarning: true; limitAfterUpdate: bigint } | { showWarning: false } {
-    const rewardStatus = useAsyncMemo(getRewardStatusLatest);
+    const rewardStatus = useAsyncMemo(getRewardStatusLatest, noOp, []);
```
orhoj, line 48: What does this have to do with the fix?

orhoj (changes requested):

app/components/PickBakerStakeAmount.tsx
shjortConcordium: I saw this and changed it to not run every time (because the dep array was undefined)

shjortConcordium:
There is still issues with the identity provider, but it seems that it is a identity-provider-service/wallet-proxy issue. :thinking: 
(Maybe the fix is to migrate to the v1 issuance)

orhoj:
> There is still issues with the identity provider, but it seems that it is a identity-provider-service/wallet-proxy issue. 🤔 (Maybe the fix is to migrate to the v1 issuance)

It is not really in scope to migrate the issuance as well. The old issuance should still work.

shjortConcordium:
> > There is still issues with the identity provider, but it seems that it is a identity-provider-service/wallet-proxy issue. 🤔 (Maybe the fix is to migrate to the v1 issuance)
> 
> It is not really in scope to migrate the issuance as well. The old issuance should still work.

Yes, it was an issue with the identity-provider-service. The fix are on stagenet.

orhoj (approved):

PR #358, opened 2024-02-28
Fix windows ledger connection issue

orhoj:
## Purpose
Fix a Ledger connectivity issue experienced on Windows.

## Changes
- Add a USB detach event listener where we check if the Ledger is actually still connected.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/app/preload/ledger/ledgerObserverImpl.ts
+++ b/app/preload/ledger/ledgerObserverImpl.ts
@@ -7,11 +7,26 @@ import type {
     Subscription,
 } from '@ledgerhq/hw-transport';
 import EventEmitter from 'events';
+import { usb } from 'usb';
 import ConcordiumLedgerClientMain from '../../features/ledger/ConcordiumLedgerClientMain';
 import { isConcordiumApp, isOutdated } from '../../components/ledger/util';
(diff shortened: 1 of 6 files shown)


app/preload/ledger/ledgerObserverImpl.ts
```diff
@@ -7,12 +7,23 @@ import type {
...
 import ledgerIpcCommands from '~/constants/ledgerIpcCommands.json';
 import { LedgerObserver } from './ledgerObserver';
 
+/**
+ * A convenience method for opening a transport by providing
+ * an empty descriptor (as the descriptor is unused downstream).
+ * @returns a promise with a {@link TransportNodeHid}
+ */
+async function openTransport(): Promise<TransportNodeHid> {
+    return TransportNodeHid.open('');
```
shjortConcordium, line 24: This has started throwing errors (assumeditably because of dependency update), could we handle those?

shjortConcordium (approved):

app/preload/ledger/ledgerObserverImpl.ts
orhoj: I have refactored a bit to better catch the errors that may be thrown.

shjortConcordium (approved):

PR #360, opened 2024-03-20
Add pyenv local version

soerenbf:
Adds the required python version for building this for pyenv users.


scripts/fileDigest.js
```diff
@@ -73,7 +73,7 @@ const { argv } = yargs
 
 const {
     file: inputFile,
-    key: privateKeyPath,
+    pkey: privateKeyPath,
```
soerenbf, line 76: This change is needed as `key` did not work anymore (maybe reserved keyword).

Ivan-Mahda (approved):

Issue #364, opened 2024-05-28
Rename anon* and shielding functionality


limemloh:
~This might include the Ledger app~

**Terminology updates**

* `anonymity revoker` = `identity disclosure authority`
* `anonymity revoker threshold` = `identity disclosure threshold`
* `anonymity revocation` = `disclosing an identity`

**Sub tasks**

- [x] Remove anon\* terms and replace as discussed
- [x] Remove shield funds button from UI
- [x] Remove shielded transfer button from UI

PR #369, opened 2024-06-06
Remove anon word

DOBEN:
## Purpose

addresses #364 

## Changes

- Remove `shielded` button.
- Remove `encryptedTransfer` button.
- Rename `anonymity revokers` to `identity disclosure authorities`.


app/pages/Accounts/AccountViewActions/AccountViewActions.tsx
```diff
@@ -36,12 +33,12 @@ const receiveAction: ActionObject = {
     height: 22,
     isDisabled: () => false,
 };
-const shieldedActions: ActionObject[] = [
+
+const unshieldedActions: ActionObject[] = [
```
soerenbf, line 37: Maybe just 
```suggestion
const actions: ActionObject[] = [
```

As these don't change anymore regardless of which balance is being displayed

soerenbf (commented):
I did a quick search, and in `app/components/ledger/CredentialInfoLedgerDetails` it states "Anonymity revocation threshold". Did we not agree to change this? I remember a discussion on AR => "authorities required", but maybe this was not approved?

limemloh (approved):

soerenbf (approved):

PR #381, opened 2025-01-31
Update web sdk

soerenbf:
In preparation for implementation of features related to protocol version 8

PR #382, opened 2025-02-06
P8/suspension status

soerenbf:
## Purpose

Show the suspension status on delegation/validation overview for accounts + show a notification when the target pool of a validator/delegator account is suspended.

Dependant on #381 being merged.

## Checklist

      hard-to-understand areas.


app/features/AccountSlice.ts
```diff
@@ -427,6 +455,35 @@ async function updateAccountFromAccountInfo(
...
+                })
+            );
+        } else if (poolStatus.currentPaydayStatus?.isPrimedForSuspension) {
+            dispatch(
+                setSuspensionStatus({
+                    address: account.address,
+                    status: 'primed',
+                })
+            );
+        }
```
limemloh, line 485: Maybe not a big issue, but from my understanding the validator can technically be both suspended and primed for suspension at the same time.

limemloh (approved):
LGTM

app/features/AccountSlice.ts
limemloh: Yes, to my understanding, being flagged 'primed for suspension' is separate from the actual suspension state.
Meaning you could be 'primed for suspension' and then self-suspend and unsuspend again, resulting in still being primed for suspension.

PR #393, opened 2025-03-04
Update node version in release workflow

limemloh:
## Purpose

Update node version in release workflow, hopefully fixing it :crossed_fingers:


andynelson (approved):
looks good, let's see how the release pipeline goes🤞

PR #400, opened 2025-03-15
Use blooo ledger library in identity issuance

limemloh:


limemloh:
The fix will be made backwards compatible on the ledger app side

PR #411, opened 2025-03-28
Fix debug-build pipeline

soerenbf:
## Purpose

Fixes the debug-build pipeline


app/app.html
soerenbf, line 1: This is not related to this, but when testing the build produced, this made no difference and is much simpler to understand.

limemloh (approved):

PR #412, opened 2025-04-23
Fix blank screen on launch and Fix auto update url

msinghster:
## Purpose

This PR addresses two separate issues related to the desktop wallet’s startup behavior and update mechanism.

**CRE-62: Blank screen on launch in DW 1.8.0**

- Issue: On Windows 11 23H2, the wallet sometimes displays a blank/grey screen on startup due to a race condition with the preload script (window.database not ready).
- Fix: Introduced a utility function waitForPreloadReady, which polls for the availability of window.database with a timeout mechanism to avoid rendering the UI prematurely.
- Tested on: Windows 11 23H2

**CRE-61: Shows error in terminal when failing to find the latest release**

- Issue: When launched from the terminal, the wallet logs a 404 error trying to fetch latest.yml due to a mismatch in the expected tag format.
- Fix: Added "vPrefixedTagName": false in package.json to align with the actual GitHub tag naming convention. This prevents 404 errors and allows update checks to function as intended.
- Behaviour: Wallet functions normally and notifies the user if updates are available in future sessions.

## Changes

- Introduced waitForPreloadReady() to handle race condition with window.database (CRE-62)
- Updated package.json to set "vPrefixedTagName": false (CRE-61)

## Checklist

- [x] I have documented my code, in particular the intent of the hard-to-understand areas

## CLA acceptance

By submitting the contribution I accept the terms and conditions of the
Contributor License Agreement v1.0
- link: https://developers.concordium.com/CLAs/Contributor-License-Agreement-v1.0.pdf

- [x] I accept the above linked CLA.


app/utils/preloadReady.ts
```diff
@@ -0,0 +1,21 @@
+export default async function waitForPreloadReady(timeout = 5000, interval = 50): Promise<void> {
```
soerenbf, line 1: Great! While it annoys me that this is necessary (as the preload script is supposed to be available as per the election docs) it seems like a good way to solve the race condition combined with the top level `await`.

I think it would be good to add docstring for this function explaining why it's there:
```suggestion
/**
 * This function is intended to be used by modules using `window.database` to interact with the database API in 
 * the preload script. This solves a race condition where this object is sometimes not available when the renderer 
 * script is executed. The problem has been observed on windows specifically with electron version 12 and 
 * election-builder version 23.
 * 
 * @returns A promise which resolves when `window.database` has been populated.
 * 
 * @example
 * await waitForPreloadReady();
 * ...
 */
export default async function waitForPreloadReady(timeout = 5000, interval = 50): Promise<void> {
```

soerenbf (approved):
Good job hunting this down 😄 Please see my suggestion for the explanatory comment on the added function.

We also need to update the changelog with the fix here.

It seems like CI is failing due to trying to build with an ubuntu version that has been removed. Could you update this as well please? 🙏

app/database/IdentityDao.ts
```diff
@@ -1,20 +1,43 @@
...
+        getNextIdentityNumber,
+        insert: insertIdentity,
+        update: updateIdentity,
+        getIdentitiesForWallet,
+        insertPendingIdentityAndInitialAccount,
+        rejectIdentityAndInitialAccount,
+        removeIdentityAndInitialAccount,
+        confirmIdentity,
+    } = window.database.identity);
+})();
```
soerenbf, line 32: Is this still required after adding the `jest.mock`? It looks very confusing. If it is still required, I think we should add a comment describing why it has to be run this way.

msinghster: Yes. The jest.mock is only used to resolve the issues with noderequest.ts file used in test/utils/transactionCosts.test.ts

msinghster: Shall I add following message on all the files that we have changed

/**
 * NOTE: These functions are loaded from the preload script, but they are not available right away.
 *
 * We have to wait for the preload to finish before we can use `window.database`.
 * However, if we use `await` at the top level, TypeScript throws the following error during Jest tests:
 *
 *   error TS1378: Top-level 'await' expressions are only allowed when the 'module' option is set to
 *   'es2022', 'esnext', 'system', 'node16', 'nodenext', or 'preserve', and the 'target' option is set to 'es2017' or higher.
 *
 * Even though our `tsconfig.json` already includes:
 *
 *     "target": "ES2020",
 *     "module": "ESNext",
 *
 * this error still appears when running tests
 *
 * To avoid changing the whole app setup or Jest environment, we use this workaround:
 * 
 * 1. Declare the variables (`let ...`)
 * 2. Wait for preload inside an async IIFE, then assign the values
 * 3. Export the variables so other parts of the app can use them
 *
 * This ensures we don’t access uninitialized APIs and keeps both the app and tests working properly.
 */

soerenbf (approved):

PR #419, opened 2025-08-03
Add create plt flow (COR-1625 and CRE-391)

DOBEN:
## Purpose

Related to: https://github.com/Concordium/concordium-node-sdk-js/pull/467
Related to: https://github.com/Concordium/concordium-node-sdk-js/pull/468
Related to: https://github.com/Concordium/concordium-ledger-app/pull/105

`createPltProposal.json` export (The `initializationParameters` are human-readable in below output and will be cbor encoded by the wallet before sending them to the ledger):
```
{
    "header": {
        "effectiveTime": {
            "@type": "bigint",
            "value": "0"
        },
        "sequenceNumber": {
            "@type": "bigint",
            "value": "1"
        },
        "timeout": {
            "@type": "bigint",
            "value": "1755032004"
        }
    },
    "payload": {
        "tokenId": {
            "@type": "tokenId",
            "value": "ETH"
        },
        "moduleRef": {
            "@type": "tokenModuleReference",
            "value": "5c5c2645db84a7026d78f2501740f60a8ccb8fae5c166dc2428077fd9a699a4a"
        },
        "decimals": 3,
        "initializationParameters": {
            "name": "name",
            "initialSupply": {
                "@type": "tokenAmount",
                "value": {
                    "value": "3000",
                    "decimals": 3
                }
            },
            "metadata": {
                "@type": "tokenMetadataUrl",
                "value": {
                    "url": "www.token.com",
                    "checksumSha256": "5c5c2645db84a7026d78f2501740f60a8ccb8fae5c166dc2428077fd9a699a4a"
                }
            },
            "governanceAccount": {
                "@type": "tokenHolder",
                "value": {
                    "type": "account",
                    "address": "4FmiTW2L2AccyR9VjzsnpWFSAcohXWf7Vf797i36y526mqiEcp",
                    "coinInfo": 919
                }
            },
            "mintable": false,
            "burnable": false,
            "allowList": true,
            "denyList": false
        }
    },
    "signatures": [],
    "type": 26
}
```

Screen to input the values:
Screenshot from 2025-08-12 23-48-53

Summary screen before signing the tx or exporting the `createPltProposal.json` file:
Screenshot from 2025-08-12 23-49-10

### DIFF
--- a/app/features/TransactionSlice.ts
+++ b/app/features/TransactionSlice.ts
@@ -122,6 +122,7 @@ function shieldedOnlyFilter(filter: TransactionFilter): TransactionFilter {
         configureDelegation: false,
         paydayAccountReward: false,
         failed: false,
+        tokenUpdate: false
     };
 
     return {
--- a/app/features/TransactionSlice.ts
+++ b/app/features/TransactionSlice.ts
@@ -122,7 +122,7 @@ function shieldedOnlyFilter(filter: TransactionFilter): TransactionFilter {
         configureDelegation: false,
         paydayAccountReward: false,
         failed: false,
-        tokenUpdate: false
+        tokenUpdate: false,
     };
 
     return {
--- a/app/features/ledger/ConcordiumLedgerClientMain.ts
+++ b/app/features/ledger/ConcordiumLedgerClientMain.ts
@@ -451,7 +451,7 @@ export default class ConcordiumLedgerClientMain {
     }
 
     signCreatePlt(
-        transaction: string, // UpdateInstruction<CreatePLTPayload>
+        transaction: string,
         serializedPayload: Buffer,
         path: number[]
     ): Promise<Buffer> {
--- a/app/features/ledger/ConcordiumLedgerClientMain.ts
+++ b/app/features/ledger/ConcordiumLedgerClientMain.ts
@@ -49,6 +49,7 @@ import {
     FinalizationCommitteeParameters,
     MinBlockTime,
     TimeoutParameters,
+    CreatePLTPayload,
 } from '~/utils/types';
 import { AccountPathInput, getAccountPath } from './Path';
 import getAppAndVersion, { AppAndVersion } from './GetAppAndVersion';
@@ -62,7 +63,6 @@ import signAddAnonymityRevokerTransaction from './SignAddAnonymityRevoker';
 import signPoolParameters from './SignPoolParameters';
 import EmulatorTransport from './EmulatorTransport';
 import verifyAddress from './verifyAddress';
-import { CreatePLTPayload } from '../../utils/types';
 
 /**
  * Concordium Ledger API.
--- a/app/features/ledger/ConcordiumLedgerClientMain.ts
+++ b/app/features/ledger/ConcordiumLedgerClientMain.ts
@@ -1,5 +1,4 @@
 import { ValidatorScoreParameters } from '@concordium/web-sdk';
-import { CreatePLTPayload } from '@concordium/web-sdk/plt';
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import type HwTransport from '@ledgerhq/hw-transport';
 import { Buffer } from 'buffer/';
@@ -63,6 +62,7 @@ import signAddAnonymityRevokerTransaction from './SignAddAnonymityRevoker';
 import signPoolParameters from './SignPoolParameters';
 import EmulatorTransport from './EmulatorTransport';
 import verifyAddress from './verifyAddress';
+import { CreatePLTPayload } from '../../utils/types';
 
 /**
  * Concordium Ledger API.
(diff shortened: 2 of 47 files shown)


app/pages/multisig/updates/CreatePLT/CreatePltParameters.tsx
```diff
@@ -0,0 +1,216 @@
...
+            />
+            <Form.Input
+                className="body2 mB20"
+                name={fieldNames.governanceAccount}
+                label={`${fieldDisplays.governanceAccount}`}
+                placeholder="4BTFaHx8CioLi8Xe7YiimpAK1oQMkbx5Wj6B8N7d7NXgmLvEZs"
+                rules={{
+                    required: requiredMessage(fieldDisplays.governanceAccount),
+                    validate: (governanceAccount: string) => {
+                        // TODO: check that governance account exists on-chain
```
soerenbf, line 141: Can we eliminate this todo? async validation is supported, so it should not be an issue to check.

app/pages/multisig/updates/CreatePLT/CreatePltParameters.tsx
```diff
@@ -0,0 +1,216 @@
...
+                rules={{
+                    required: requiredMessage(fieldDisplays.tokenId),
+                    validate: (value: string) => {
+                        const cleaned = value.replace(/[^a-zA-Z0-9.%-]/g, '');
+                        if (cleaned.length > 128) {
+                            return 'Must be 128 characters or less.';
+                        }
+                        if (value !== cleaned) {
+                            return "Only letters, numbers, '.', '%', and '-' are allowed.";
+                        }
```
soerenbf, line 77: I would flip these:

```suggestion
                        if (value !== cleaned) {
                            return "Only letters, numbers, '.', '%', and '-' are allowed.";
                        }
                        if (cleaned.length > 128) {
                            return 'Must be 128 characters or less.';
                        }
```

otherwise you're checking the length the value that is not in the input field, which seems weird to me

app/pages/multisig/updates/CreatePLT/CreatePltParametersView.tsx
```diff
@@ -0,0 +1,102 @@
...
+                <div className="body3 mono mB20">
+                    {initializationParameters.mintable?.toString() ?? 'False'}
+                </div>
+            </div>
+            <div>
+                <Label className="mB5">{fieldDisplays.burnable}:</Label>
+                <div className="body3 mono mB20">
+                    {initializationParameters.burnable?.toString() ?? 'False'}
+                </div>
+            </div>
```
soerenbf, line 99: There's so much data to represent here, that the view of the proposal does not look very nice in the proposal overview. I don't know if it warrants changing it.

If you do want to change it, you'd need to update it in `UpdateInstructionDetails.module.scss`. I found that `25px` between them looks nice 🤷

app/features/ledger/ConcordiumLedgerClientMain.ts
```diff
@@ -448,6 +449,20 @@ export default class ConcordiumLedgerClientMain {
...
+        path: number[]
+    ): Promise<Buffer> {
+        return signUpdateTransaction(
+            this.transport,
+            0x48,
+            path,
+            transaction,
+            serializedPayload
+        );
+    }
```
soerenbf, line 464: This will not work, the data needs to be sent in batches. Have a look at the relevant doc for the instruction: https://github.com/Concordium/concordium-ledger-app/blob/b371047440bfb8dca8b3dd5100e3f15e44c9d134/governance-app/doc/ins_create_plt.md

app/pages/multisig/updates/CreatePLT/CreatePltParameters.tsx
DOBEN: Once we finalize the exact validation to do in `concordium-client`, the suggestion would be to do the same validation in the desktop wallet. Seems that likely we would only need to add the check if the governance account exists on-chain to DW hence that commnet. 

https://linear.app/concordium/issue/COR-1645/concordium-client-validation-improvements-add-better-error-reporting

app/features/ledger/ConcordiumLedgerClientMain.ts
soerenbf: You can also have a look at the e2e test, might be easier to see how the data is transferred there:

https://github.com/Concordium/concordium-ledger-app/blob/b371047440bfb8dca8b3dd5100e3f15e44c9d134/tests/src/governance/createPlt.test.ts

app/features/TransactionSlice.ts
```diff
@@ -122,6 +122,7 @@ function shieldedOnlyFilter(filter: TransactionFilter): TransactionFilter {
         configureDelegation: false,
         paydayAccountReward: false,
         failed: false,
+        tokenUpdate: false,
```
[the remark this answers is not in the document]
soerenbf: ah, of course 😄

msinghster (approved):

PR #422, opened 2025-08-26
Support v2 authorizations (COR-1766)

DOBEN:
## Purpose

Related to: https://github.com/Concordium/concordium-ledger-app/pull/116

## Changes

- Fix display of `new size of level 2 key set` which now accounts for removed keys correctly
- Update the flow for signing authorizations updates with support for v2 authorizations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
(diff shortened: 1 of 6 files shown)


soerenbf (approved):
Looks good and seems to work 👍

PR #423, opened 2025-08-27
Fix: Desktop Wallet Auto-Update & Multi-Account support

msinghster:
## Purpose

This PR addresses following issues and improvements in the Desktop Wallet:

- Fix the auto-update functionality.
- Fix second account creation.

## Changes

Fix the auto-update functionality.
- Adjusted verify.ts to correctly match the downloaded file path during update verification.

Fix second account creation
- Updated package.json file to include V6 SDK to use the following functions during 2nd account creation, as there is a difference in the way getCredentialDeploymentTransactionHash and createCredentialDeploymentTransaction handle the values of CredentialDeploymentTransaction.
- The changes done in CredentialDeploymentTransaction were of additional attribute keys. In the new SDK, additional keys were introduced. We have types.ts that exclusively defines the keys to be used in identity creation. They are similar to V6 SDK keys.

## Checklist

- [ ] I have documented my code, in particular the intent of the
      hard-to-understand areas.
- [x] I have updated the CHANGELOG.

## CLA acceptance

By submitting the contribution I accept the terms and conditions of the Contributor License Agreement v1.0
- link: https://developers.concordium.com/CLAs/Contributor-License-Agreement-v1.0.pdf

- [x] I accept the above linked CLA.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,6 @@
 ### Fixed
 
 - Unable to add a 2nd account in a DW
-- Update wallet proxy integration to v2 accTransactions
 - Auto update not working for Desktop wallet
 
 ## 1.8.2


app/preload/preloadTypes.ts
```diff
@@ -140,6 +141,12 @@ export type GRPC = {
         transaction: CredentialDeploymentTransaction,
         signatures: string[]
     ) => Promise<string>;
+
+    /* sendCredentialDeploymentTransactionV6: (
+        transaction: CredentialDeploymentTransactionv6,
+        signatures: string[]
+    ) => Promise<string>; */
+
```
soerenbf, line 149: Is this still needed? If it is, please add a comment describing when it should be reintroduced, but otherwise please remove 😊

app/preload/http.ts
```diff
@@ -108,7 +108,7 @@ async function getTransactions(
         filters += `&blockTimeTo=${timestamp}`;
     }
 
-    let proxyPath = `/v1/accTransactions/${address}?limit=${limitResults}&includeRawRejectReason${filters}&order=${order.toString()}`;
+    let proxyPath = `/v2/accTransactions/${address}?limit=${limitResults}&includeRawRejectReason${filters}&order=${order.toString()}`;
```
soerenbf, line 111: Please check the [changes introduced with the v2 endpoint](https://github.com/Concordium/concordium-wallet-proxy/blob/main/README.md#details-required). We need to be able to handle the two new types introduced, i.e.

```
{type: 'validatorPrimedForSuspension', ...}
{type: 'validatorSuspended', ...}
```

I don't know if there's a fallback mechanism in place or not, but the point of the upgrade was to show this as part of the activity log on an account.

msinghster: We can ignore this update for now as I have only changed the V2 endpoint

app/preload/preloadTypes.ts
msinghster: Not required. so removed

app/preload/http.ts
msinghster: No fallback is there. but we are passing variable to filter for encryptedAmountTransferWithMemo and encryptedAmountTransfer

